PR #4, opened 2018-05-14
feat(sms): switch to direct Telstra client

jonathaningram:
Keep the Identity client in place and fallback to it if the Telstra
client fails.

Note: the env vars changed in order to make it clear to which SMS client
they apply.

Note: the Telstra client doesn't account for provisioning the number -
i.e. it assumes that it has been provisioned separately.

### DIFF
--- a/api/app/clients/sms/telstra.py
+++ b/api/app/clients/sms/telstra.py
@@ -1,0 +1,59 @@
+from monotonic import monotonic
+from app.clients.sms import SmsClient
+import Telstra_Messaging
+from Telstra_Messaging.rest import ApiException
+
+class TelstraSMSClient(SmsClient):
+    def __init__(self, client_id=None, client_secret=None, *args, **kwargs):
+        super(TelstraSMSClient, self).__init__(*args, **kwargs)
+        self._client_id = client_id
+        self._client_secret = client_secret
+
+    def init_app(self, app, *args, **kwargs):
+        self.current_app = app
+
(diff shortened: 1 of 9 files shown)


api/app/clients/sms/telstra.py
```diff
@@ -0,0 +1,59 @@
...
+    def auth(self):
+        grant_type = 'client_credentials'
+
+        api_instance = Telstra_Messaging.AuthenticationApi()
+
+        start_time = monotonic()
+        try:
+            resp = api_instance.auth_token(self._client_id, self._client_secret, grant_type)
+
+            self.current_app.logger.error("Telstra auth request succeeded")
```
dagoof: logger.info

api/app/clients/sms/telstra.py
```diff
@@ -0,0 +1,59 @@
...
+        return 'telstra'
+
+    def get_name(self):
+        return self.name
+
+    def send_sms(self, to, content, reference, sender=None):
+        conf = self.auth()
+
+        api_instance = Telstra_Messaging.MessagingApi(Telstra_Messaging.ApiClient(conf))
+        payload = Telstra_Messaging.SendSMSRequest(to=to, body=content)
```
dagoof: we don't need it now, but does telstra let us set a sender?

jonathaningram:
Resolved by https://github.com/govau/notifications/pull/11

PR #16, opened 2018-07-19
update status of sent SMTP notification

jonathaningram:


api/app/delivery/send_to_providers.py
```diff
@@ -134,7 +134,7 @@ def send_email_to_provider(notification):
                 reply_to_address=validate_and_format_email_address(email_reply_to) if email_reply_to else None,
             )
             notification.reference = reference
-            update_notification(notification, provider)
+            update_notification(notification, provider, status=status)
```
dagoof: is this an oversight/bug that exists or is it a hack to take advantage of a convenient codepath?

jonathaningram: Yeah base class has no return type so I figured this should be fine.

I'm pushing an update to this now that I understand it better too.

PR #18, opened 2018-07-26
feat(makefile): move apply-services into makefile

dagoof:


Makefile
```diff
@@ -1,16 +1,22 @@
...
-DEPLOY_ENV ?= notifications
-CF_API     ?= https://api.system.$(CLD_HOST)
-CF_ORG     ?= dta
-CF_SPACE   ?= $(DEPLOY_ENV)
-CF_HOME    ?= $(HOME)
+CLD_HOST    ?= y.cld.gov.au
+DEPLOY_ENV  ?= notifications
+CF_API      ?= https://api.system.$(CLD_HOST)
+CF_ORG      ?= dta
+CF_SPACE    ?= $(DEPLOY_ENV)
```
jonathaningram: @dagoof you can remove `DEPLOY_ENV` and just make `CF_SPACE= notifications`. Unless you were planning on using `DEPLOY_ENV` for other things later?

ci/apply-services
```diff
@@ -1,24 +0,0 @@
-#!/usr/bin/env bash
```
jonathaningram: @dagoof you happy to keep the .sample.json files in or is there a better way?

dagoof: i want to leave them in. we need a better solution eventually but they're pretty crucial right now so we at least know what should exist

Makefile
dagoof: I think I wanted to decouple our notion of environment /app name from the deploy mechanism. but I don't know if it even matters anymore

PR #19, opened 2018-07-27
ci: add beginnings of b-cld deployment

jonathaningram:
- [x] manifest files are just copy-pasted between y and b. could inherit from base and/or move env vars to UPSs
- [x] after y-cld successfully deploys, it automatically starts deploying the 3 b-cld apps. should this also have an approval before it starts?
- [x] not sure what the difference is between the 3 deploys and the final deploy of admin when the approval is given. I feel like this is wrong.


.circleci/config.yml
```diff
@@ -12,106 +20,156 @@ references:
...
-  deploy-celery:
-    docker:
-      - image: circleci/python:3.6.5-node
-    steps:
-        - *cf_install
-        - attach_workspace:
-            at: .
-        - *cf_login
-        - run: make -C api deploy-celery
+    <<: *deploy_admin_y_cld
```
dagoof: can we collapse these and put environment vars in the workspace part? 

like: 

* deploy defaults, -- deploy-api, deploy-admin, deploy-celery (staging)
* manual unlock
* set environment and re-perform deploys. deploy-api deploy-admin deploy-celery

dagoof:
adjusted according to the two points on your checklist

PR #24, opened 2018-08-07
feat(homepage): replace govuk svg with dta svg

dagoof:


admin/app/templates/views/signedout.html
```diff
@@ -99,83 +99,6 @@ <h2 class="with-keyline">
       </div>
     </div>
   </div>
-  <div class="product-page-section">
```
jonathaningram: @dagoof can probably remove some CSS if you want, but no big deal.

PR #30, opened 2018-08-14
fix(copy): adjust a misplaced colon

dagoof:


admin/app/main/views/service_settings.py
```diff
@@ -476,7 +476,7 @@ def service_set_sms_prefix(service_id):
         'on' if current_service['prefix_sms'] else 'off'
     ))
 
-    form.enabled.label.text = 'Start all text messages with ‘{}:’'.format(current_service['name'])
+    form.enabled.label.text = 'Start all text messages with ‘{}’:'.format(current_service['name'])
```
jonathaningram: @dagoof how did you test this change?

dagoof: I have written an integration test suite around this feature that will be enabled in an upcoming PR

jonathaningram: For those listening at home, the answer we were looking for was "colonoscopy"...

PR #33, opened 2018-08-20
Use OS's SQLALCHEMY_DATABASE_URI development environment

andrewoh531:


.gitignore
```diff
@@ -1,3 +1,6 @@
 */vendor
 */requirements.txt
 */pipenv-requirements.txt
+
+.idea/
```
jonathaningram: @andrewoh531 this is fine, but you might like to use a personal gitignore instead otherwise you'll have to make this change in all repos that you work with: https://www.atlassian.com/git/tutorials/saving-changes/gitignore#personal-git-ignore-rules

andrewoh531: Can do next time :)

PR #35, opened 2018-08-22
feat-api-wrappers

dagoof:
Adds two new projects. rpc-api and graphql-api

These were the output of our api-expedition that have been restructured into root-level deployable apps.

they both deploy to feature branches, dev, stg, but not prod. They're so far away from being prod ready that I don't think it's appropriate to go there yet.

I will squash all of these commits when it comes time to merge


rpc-api/cmd/notify-rpc-api/grpc.go
```diff
@@ -0,0 +1,447 @@
...
+func (c Client) Do(req *http.Request) (*http.Response, error) {
+	token, err := CreateJWT(c.ServiceID, c.API_Key)
+	if err != nil {
+		return nil, err
+	}
+
+	req.Header.Add("Content-type", "application/json")
+	req.Header.Add("Authorization", fmt.Sprintf("Bearer %s", token))
+	req.Header.Add("X-Custom-Forwarder", "")
+	req.Header.Add("User-agent", "NOTIFY-API-PYTHON-CLIENT/4.9.0")
```
jonathaningram: Did you mean for this to be the Python user agent?

rpc-api/cmd/notify-rpc-api/grpc.go
```diff
@@ -0,0 +1,447 @@
...
+func (c Client) Post(path string, body io.Reader) (*http.Response, error) {
+	req, err := http.NewRequest("POST", path, body)
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := c.Do(req)
+	if err != nil {
+		return nil, err
+	}
```
jonathaningram: `defer resp.Body.Close()` (and elsewhere)

rpc-api/cmd/notify-rpc-api/grpc.go
```diff
@@ -0,0 +1,447 @@
...
+
+	return JSONDataNested(resp.Body)
+}
+
+func (c Client) CreateTemplate(serviceID string, template Template) (json.RawMessage, error) {
+	var buf bytes.Buffer
+	err := json.NewEncoder(&buf).Encode(template)
```
jonathaningram: Check `err`

rpc-api/cmd/notify-rpc-api/grpc.go
```diff
@@ -0,0 +1,447 @@
...
+	"fmt"
+	"io"
+	"log"
+	"net"
+	"net/http"
+	"net/url"
+	"os"
+	"time"
+
+	"github.com/govau/notify/rpc-api/notify"
```
jonathaningram: Move to end and alias as `pb`:

```go
import (
  "bytes"

  "google.golang.org/grpc"

  pb "github.com/govau/notify/rpc-api/notify"
)
```

This is so that you know when you see `notify.Service` that you are referring to proto code. Especially important if one uses a `notify` package at the domain level because otherwise you'd have a conflict.

E.g. https://github.com/AusDTO/tell-us-once/blob/master/tuo/coci/order.go

rpc-api/cmd/notify-rpc-api/grpc.go
```diff
@@ -0,0 +1,447 @@
...
+		return "", err
+	}
+
+	cl := jwt.Claims{
+		Issuer:   clientID,
+		IssuedAt: jwt.NewNumericDate(time.Now()),
+	}
+
+	return jwt.Signed(sig).Claims(cl).CompactSerialize()
+	return jwt.Signed(sig).Claims(cl).FullSerialize()
```
jonathaningram: Two returns.

rpc-api/cmd/notify-rpc-api/grpc.go
```diff
@@ -0,0 +1,447 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := c.Do(req)
+	if err != nil {
+		return nil, err
+	}
+
+	if resp.StatusCode > 300 {
```
jonathaningram: Did you mean `>= 300`?

rpc-api/cmd/notify-rpc-api/grpc.go
```diff
@@ -0,0 +1,447 @@
...
+		return resp, errors.New("bad status code")
```
jonathaningram: Should we do `return resp, fmt.Errorf("bad status code %d", resp.StatusCode)` so that it's a bit easier to debug issues?

rpc-api/cmd/notify-rpc-api/grpc.go
```diff
@@ -0,0 +1,447 @@
...
+		Permissions:      service.Permissions,
+		Users:            service.Users,
+	}
+
+}
+
+type Services []Service
+
+func (s Services) Protoify() *notify.Services {
+	services := [](*notify.Service){}
```
jonathaningram: I didn't know you could do this. I usually just do `make([]*notify.Service, 0)` to get around the JSON `null / []` issue. No need to change, just didn't know.

rpc-api/cmd/notify-rpc-api/grpc.go
```diff
@@ -0,0 +1,447 @@
...
+
+	for _, service := range s {
```
jonathaningram: Per my comment above making the `notify` package import `pb`, this would now look like this:

```go
servicesp := [](*pb.Service){}

for _, s := range services {
  servicesp = append(servicesp, s.Protoify())
}
```

The reason for this is because looking at the method, it's unclear out of variables `s`, `services` and `service` what is proto and what is not. Suffixing with `p` solves this.

rpc-api/cmd/notify-rpc-api/grpc.go
```diff
@@ -0,0 +1,447 @@
...
+		UserToService:    service.UserToService,
+		Permissions:      service.Permissions,
+		Users:            service.Users,
+	}
+
+}
+
+type Services []Service
+
+func (s Services) Protoify() *notify.Services {
```
jonathaningram: Instead of `Protoify` just `Proto`.

rpc-api/cmd/notify-rpc-api/grpc.go
```diff
@@ -0,0 +1,447 @@
...
+	return templates.Protoify(), err
+}
+
+func (s *server) CreateTemplate(ctx context.Context, in *notify.CreateTemplateRequest) (*notify.Template, error) {
+
+	templateJSON, err := s.c.CreateTemplate(
+		in.ServiceId,
+		Template{
+			Name:         in.Name,
+			TemplateType: "email",
```
jonathaningram: Is this always `email`?

rpc-api/cmd/notify-rpc-api/grpc.go
```diff
@@ -0,0 +1,447 @@
...
+			Subject:      in.Subject,
+			Content:      in.Content,
+			Service:      in.ServiceId,
+			CreatedBy:    in.CreatedBy,
+			CreatedAt:    time.Now().Format(time.RFC3339),
+			ProcessType:  "normal",
```
jonathaningram: Does this need to be extracted to a const somewhere?

rpc-api/cmd/notify-rpc-api/grpc.go
```diff
@@ -0,0 +1,447 @@
...
+	jose "gopkg.in/square/go-jose.v2"
+	"gopkg.in/square/go-jose.v2/jwt"
+)
+
+type Client struct {
+	http.Client
+
+	BaseURL     *url.URL
+	ServiceID   string
+	API_Key     string
```
jonathaningram: `APIKey`

proto/notify.proto
```diff
@@ -0,0 +1,93 @@
+syntax = "proto3";
+
+service Notify {
+  rpc GetUser (ForUserRequest) returns (User) {}
```
jonathaningram: Each `rpc` should have their input as the name of the RPC plus `Request` and their output + `Response` even if it's empty for now e.g.:

```
rpc GetUser (GetUserRequest) returns (GetUserResponse) {}
```

E.g. see https://gophers.slack.com/archives/C8T4Y2248/p1533921988000302?thread_ts=1533920932.000058&cid=C8T4Y2248

> Good protobuf practice is to return something, even an empty DeleteResponse{}. The reasoning being a change from empty to a message is a breaking change, but a change to add a field is a backwards compatible change.
> Having done many years with pb in production environments, I've learned the hard way to follow that advice. :slightly_smiling_face:

rpc-api/cmd/notify-rpc-api/grpc.go
dagoof: this is prototype code

rpc-api/cmd/notify-rpc-api/grpc.go
dagoof: this is prototype code

rpc-api/cmd/notify-rpc-api/grpc.go
dagoof: if we create a notify package we can rename this. but we won't, because we'll be deleting this entire file.

rpc-api/cmd/notify-rpc-api/grpc.go
dagoof: dont know what the server expects. this was lifted directly from python client api to get it working. I don't know what this is meant to be or if behaviour changes

proto/notify.proto
dagoof: that post doesn't say anything about naming conventions. all of these rpc calls have a request/return message

jonathaningram:
```
rpc GetUsers (Empty) returns (Users) {}
```

If `Empty` is used by any other `rpc`, then this is a problem. E.g. now you want to add some fields to the `GetUsers` call like `bool hide_inactive`, `int64 limit`. To do that they have to be added to `Empty` which doesn't make sense if other calls are also using `Empty`. So you then change the input to `GetUsersRequest` and now have a breaking change. So it's better to just use the Request/Response rule from the beginning even if it's a tiny bit more upfront work.

For the record, I had a quick look in GCP's repo and they have some cases where they use the "pragmatic" way like you have done:

```
rpc GetService (ServiceRequest) returns (Service) {}
```

However, you can't add anything to `Service` without affecting other things, so it's safest to just use `GetServiceRequest` and `GetServiceResponse`.

jonathaningram:
> your feedback is noted and valid but I want it to be clear that none of this is production ready; nor was that the scope of this ticket. this is about scaffolding out a space for us to start working on these services.
> 
> this grpc stuff is clearly all going to get deleted because we can't even run grpc on our infra. it makes no sense to go through it with a fine toothed comb

No worries. I worry about committing some of the code if it's just going to be deleted late thoughr? Would it be better to just change the notify.proto to include just one RPC in order to actually prove that the servers are hooked up. i.e. essentially just a hello world call that is obvious we will replace later.

We should still fix the few issues like double returns, missing err check, defer resp.Body.Close before committing.

Regarding the gRPC stuff, we should talk about that and decide on a plan, but it probably just means using Twirp instead of gRPC as we've done in TUO.

dagoof:
Agree. there's some stuff that's obviously wrong here (including the fact that it's pointing to `localhost`). err check etc. I've fixed those in a commit.

I really just intend this to be a place where we can eventually have a functioning Go app without needing to deal with the build and deploy at that point.

jonathaningram:
> Agree. there's some stuff that's obviously wrong here (including the fact that it's pointing to localhost). err check etc. I've fixed those in a commit.
> 
> I really just intend this to be a place where we can eventually have a functioning Go app without needing to deal with the build and deploy at that point.

Yep sounds good.

proto/notify.proto
dagoof: this makes sense

andrewoh531:
General questions from me:
- Does it makes sense for this to be its own repo? 
- Is it Go convention to have single large files?

dagoof:
The hope was to bundle everything together in one repo for our own sanity, but the ridiculousness of $GOPATH is pushing me to consider otherwise.

Go convention is to break files in terms of whatever makes sense. you can have as many files as you want all namespaced under one package. put things that belong together, together

sometimes that's 2500 lines https://golang.org/src/net/http/transport.go

sometimes it's 1300 https://golang.org/src/encoding/json/decode.go

sometimes much less. There's certainly no such "one class per file" idiom or anything like that.

dagoof:
tagged here. closing this

https://github.com/govau/notify/releases/tag/api-wrappers

PR #36, opened 2018-08-24
feat(two-factor): change link text, reworded

dagoof:
also got rid of misleading back arrow.

the ticket would have preferred for a larger restructure to happen but
it's just too much of a change to do at this stage.

### DIFF
--- a/admin/app/templates/views/two-factor-email.html
+++ b/admin/app/templates/views/two-factor-email.html
@@ -14,7 +14,7 @@
     <p>Clicking the link will open Notify in a new browser window, so you can close this one.</p>
     {{ page_footer(
       secondary_link=url_for('main.email_not_received'),
-      secondary_link_text="Didn't receive an email?"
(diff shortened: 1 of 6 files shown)


admin/app/templates/views/two-factor-email.html
```diff
@@ -14,7 +14,7 @@ <h1 class="heading-large">{{ title }}</h1>
     <p>Clicking the link will open Notify in a new browser window, so you can close this one.</p>
     {{ page_footer(
       secondary_link=url_for('main.email_not_received'),
-      secondary_link_text='Not received an email?'
+      secondary_link_text="Didn't receive an email?"
```
jonathaningram: It's a nit, but can you use the fancy `’` single quote to match the rest of the site’s copy 😎

dagoof: gross. yeah

PR #38, opened 2018-08-27
Remove integration to DeskPro and send email instead

andrewoh531:


api/migrations/versions/0197_creating_service_template_when_a_.py
```diff
@@ -0,0 +1,59 @@
+"""
```
jonathaningram: @andrewoh531 I wonder if you should rename the migration file. It was truncated and doesn't really explain what it was intending to do.

api/migrations/versions/0197_creating_service_template_when_a_.py
```diff
@@ -0,0 +1,59 @@
...
+down_revision = '0196'
+
+template_id = '831139ba-a984-11e8-814c-80e6501bb488'
+
+content = """Hi Notify Support,
+
+The following service has requested to go live: 
+
+Name: ((service_name)) 
+Url: ((service_url))
```
jonathaningram: @andrewoh531 and it's minor but while you are renaming this template can you use `URL` instead of `Url`.

api/migrations/versions/0197_creating_service_template_when_a_.py
andrewoh531: Sure

api/migrations/versions/0197_creating_service_template_when_a_.py
andrewoh531: Sure

dagoof:
love it. awesome.

PR #49, opened 2018-09-07
feat(logo): use new bucket, with separate creds

dagoof:
Our cloud team provisions new s3 credentials per-bucket. this makes it
impossible to access multiple buckets using a set of global creds.

Here we manage s3 connections by bundling a set of credentials together
with the bucket name they are allowed to access.

### DIFF
--- a/admin/app/main/s3_client.py
+++ b/admin/app/main/s3_client.py
@@ -1,7 +1,7 @@
 import uuid
 
 import botocore
-from boto3 import resource
+from boto3 import resource, Session
 from flask import current_app
 from notifications_utils.s3 import s3upload as utils_s3upload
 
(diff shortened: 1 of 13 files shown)


admin/app/main/s3_client.py
```diff
@@ -1,98 +1,149 @@
...
+
+    def copy_object(self, old_name, new_name):
+        self.get_object(new_name).copy_from(
+            CopySource='{}/{}'.format(self._bucket_name, old_name))
+
+
+class Clients(object):
+    @property
+    def logo(self):
+        return S3Client(
```
dagoof: this might need some micro-optimization.

s3 boto library re-uses s3 sessions instead of re-instantiating them

dagoof:
I've just half-done the cdn stuff and pointed it to `/static-logo` for now. the alternative was leaving it pointing to alphagov stuff which was even more wrong.

you couldn't even see the stuff that had just been uploaded.

admin/app/main/s3_client.py
dagoof: see here

https://github.com/boto/boto3/blob/develop/boto3/__init__.py#L72

admin/app/main/views/service_settings.py
```diff
@@ -827,9 +827,11 @@ def get_branding_as_value_and_label(email_branding):
 
 
 def get_branding_as_dict(email_branding):
+    # TODO: point this to our cdn when we make it
+
     return {
         branding['id']: {
-            'logo': 'https://{}/{}'.format(get_cdn_domain(), branding['logo']),
+            'logo': '//{}/static-logo/{}'.format(get_cdn_domain(), branding['logo']),
```
andrewoh531: I think we may need to be careful what we go "live" with and what our end solution for serving assets will be. e.g. if we end up setting up a CDN using AWS services (Route53, cloudfront) we may want to have a subdomain i.e. `static-logo.{base-url}/` rather than a relative path `{base-url}/static-logo/`

jonathaningram: i'll make a ticket.

PR #51, opened 2018-09-10
Test billing

andrewoh531:


api/tests/app/email_branding/test_rest.py
```diff
@@ -13,11 +13,16 @@ def test_get_email_branding_options(admin_request, notify_db, notify_db_session)
         'email_branding.get_email_branding_options'
     )['email_branding']
 
-    assert len(email_branding) == 2
+    assert 6 == len(email_branding)
```
jonathaningram: Any reason to change assert order? I suspect most of the codebase is using `actual, expected`, so I'd just leave these as is.

api/tests/app/email_branding/test_rest.py
```diff
@@ -13,11 +13,16 @@ def test_get_email_branding_options(admin_request, notify_db, notify_db_session)
...
     assert {
         email_branding['id'] for email_branding in email_branding
     } == {
-        str(email_branding1.id), str(email_branding2.id)
```
jonathaningram: What's the reason there are now 6 brandings, not 2?

api/tests/app/email_branding/test_rest.py
andrewoh531: Hmmm...I was getting an output like `Expected:6 Actual:2` with the expected value on the left hand side of the `==`, however seems like this is only for my IDE. I can change the order around again.

api/tests/app/email_branding/test_rest.py
andrewoh531: With the db being recreated every time, there are migrations that adds rows into the `email_branding` table: https://github.com/govau/notify/blob/f786fd0cdf4fd9c7e69608705e1cca68be4fca71/api/migrations/versions/0161_email_branding.py#L33

andrewoh531:
The `deploy-dev-api` step is failing cause it doesn't contain the latest 199 migration script:

```   2018-09-10T16:36:34.39+1000 [APP/PROC/WEB/0] ERR alembic.util.exc.CommandError: Can't locate revision identified by '0199'
   2018-09-10T16:36:34.54+1000 [APP/PROC/WEB/0] ERR 2018/09/10 06:36:34 exit status 1
   2018-09-10T16:36:34.54+1000 [APP/PROC/WEB/0] ERR make: *** [setup-db] Error 1```

PR #52, opened 2018-09-10
fix(tests): make test-delivery

jonathaningram:
To fix these tests, it was necessary to update the twilio and telstra
provider details in the database. Twilio now has a priority of 10
(highest), Telstra has a priority of 15, MMG and Firetext are both
disabled and now have a priority of 20. Twilio is also updated to
support international sending.

It is still worth removing MMG and Firetext-related code and records
since they are no longer necessary and unsupported.


api/tests/app/delivery/test_send_to_providers.py
```diff
@@ -107,26 +107,26 @@ def test_should_send_personalised_template_to_correct_email_provider_and_persist
...
+    mocker.patch('app.smtp_client.send_email', return_value=['reference', 'sent'])
 
     send_to_providers.send_email_to_provider(
         db_notification
     )
 
-    app.aws_ses_client.send_email.assert_called_once_with(
-        '"Sample service" <sample.service@test.notify.com>',
+    app.smtp_client.send_email.assert_called_once_with(
+        '"Sample service" <sample.service@notifytest.gov.au>',
```
andrewoh531: Is mail actually being sent out to this email?

jonathaningram: Nope, it's mocking it on line 110: https://github.com/govau/notify/pull/52/files#diff-c6288ac0b74b33c97c8b1882d7af6877R110 and returning fake values.

api/migrations/versions/0199_change_provider_details.py
```diff
@@ -0,0 +1,36 @@
...
+
+revision = '0199'
+down_revision = '0198'
+
+
+def upgrade():
+    op.execute("UPDATE provider_details SET supports_international=True WHERE identifier='twilio'")
+    op.execute("UPDATE provider_details_history SET supports_international=True WHERE identifier='twilio'")
+    op.execute("UPDATE provider_details SET priority=15 WHERE identifier='telstra'")
+    op.execute("UPDATE provider_details_history SET priority=15 WHERE identifier='telstra'")
```
dagoof: this history table is like a log; semantically we should create a new entry for this rather than updating existing ones.

We'll wipe out all of our admin update history (not that there is much) if we go ahead with this

It's basically the same as  "provider_details" without the unique id

api/migrations/versions/0199_change_provider_details.py
```diff
@@ -0,0 +1,36 @@
...
+from alembic import op
+import sqlalchemy as sa
+
+
+revision = '0199'
+down_revision = '0198'
+
+
+def upgrade():
+    op.execute("UPDATE provider_details SET supports_international=True WHERE identifier='twilio'")
```
dagoof: our current twilio number can only send to domestic numbers. we should look at purchasing one that supports international; or creating another "twilio-international" provider that could be used here

jonathaningram: peter and I checked and the current twilio number can send to international numbers (except for USA/Canda which is +1 prefix)

jonathaningram:
Replaced by https://github.com/govau/notify/pull/61

PR #53, opened 2018-09-10
feat(tests): use local postgres by default

dagoof:
And allow fallback to docker.

We also require database explicity in some tests that just happened to
work due to a leftover database state between tests on a non-worker
database instance. Now all tests that require database state must state
it explicity or pull a dependency that uses it. This forces them to use
a properly set-up and torn-down worker database instance; and lets us
avoid running migrate on the root instance.


api/Makefile
```diff
@@ -70,26 +66,16 @@ test: lint tests
 lint:
 	$(RUNNER) flake8 .
 
-tests: PIPENV_DOTENV_LOCATION := $(TEST_PIPENV_DOTENV_LOCATION)
 tests: version-file
-	$(MAKE) setup-test-db
```
jonathaningram: Do we still need the target `setup-test-db` now that you've remove their invocation?

.circleci/config.yml
```diff
@@ -10,7 +10,7 @@ references:
       - image: circleci/python:3.6.5-node
       - image: circleci/postgres:9.6-alpine
         environment:
-          POSTGRES_USER: postgres
+          POSTGRES_USER: circleci
```
jonathaningram: Why? Isn't `postgres` the standard name?

api/Makefile
dagoof: yes will do in a PR later as discussed

PR #54, opened 2018-09-10
feat(tests): skip tests regarding letters

dagoof:


api/tests/app/letters/test_letter_schemas.py
```diff
@@ -7,6 +7,7 @@
 from app.letters.letter_schemas import letter_job_ids
 from app.schema_validation import validate
 
+pytest.skip("we no longer support sending letters", allow_module_level=True)
```
andrewoh531: Do you think we'll be supporting letters in the future? Wonder if we should just remove these files.

jonathaningram: This implies we previously supported them. Might we worth wording it "we do not support sending letters" just so any readers are not confused.

api/tests/app/letters/test_anyways.py
```diff
@@ -0,0 +1,4 @@
+import pytest
```
jonathaningram: What's your thinking with the suffix `_anyways`? Is there a better name for it - not clear what your intention is. Is it like a `test_keep.py` or a `test_skip.py`?

PR #61, opened 2018-09-11
Test delivery, clients

jonathaningram:


api/app/__init__.py
```diff
@@ -43,6 +44,7 @@
     user=os.getenv('SMTP_USER'),
     password=os.getenv('SMTP_PASSWORD'),
 )
+aws_ses_client = AwsSesClient()
```
andrewoh531: Are we planning to use SES?

jonathaningram: Probably later, but there's a whole suite of AWS SES tests that we can keep in if we just initialise this client here.

PR #64, opened 2018-09-13
Display commit sha and build numbers properly

andrewoh531:


andrewoh531:
See the following endpoints:
- https://notify-api-f-travis.apps.y.cld.gov.au/_status
- https://notify-f-travis.apps.y.cld.gov.au/_status

PR #65, opened 2018-09-13
feat(header): COA logo with 'Aus Government' text

jonathaningram:


jonathaningram:
@dagoof @andrewoh531 not real pleased about the size and contents of the SVG (see adobe class names), but I got this off Marketplace and would rather just use theirs for now instead of having this fail the design again.

PR #77, opened 2018-09-24
Do not require auth for the static logo endpoint

andrewoh531:
Will need to test functionality against: https://notify-f-logo.apps.y.cld.gov.au


admin/app/utils.py
```diff
@@ -64,13 +64,10 @@ def authenticate():
         {'WWW-Authenticate': 'Basic realm="Login Required"'})
 
 
-def requires_auth(f):
-    @wraps(f)
-    def decorated(*args, **kwargs):
-        if not check_auth(current_app.config, request.authorization):
+def requires_auth():
+    if not check_auth(current_app.config, request.authorization):
+        if request.endpoint != 'main.static_logo':
```
dagoof: could we exit early here?

```
def requires_auth():
    if request.endpoint == 'main.static_logo':
        return True

    if not check_auth(current_app.config, request.authorization):
        return authenticate()
```

it's a small thing but it shows the intent that we always want to display static logo regardless of the auth situation

andrewoh531: Sure

PR #80, opened 2018-09-26
feat(features): change some copy

jonathaningram:
We specifically point out that branding can be customised but we will
work with the service to use their branding rather than just allowing
whatever branding the service asks for.

We don't give any commitments around sending messages yet.


jonathaningram:
Can test https://notify-f-features-page.apps.y.cld.gov.au/

PR #83, opened 2018-09-27
Include bandit static code analysis check configurations

andrewoh531:
Configurations in `.circleci/config.yml` and `Makefile` included but not included in the workflow.


status/Makefile
```diff
@@ -47,4 +47,5 @@ create-service-mysql:
 
 clean:
 
-.PHONY: setup setup-dev run deploy create-service-mysql setup-db migrate-db help clean
+
+.PHONY: setup setup-dev run deploy create-service-mysql setup-db migrate-db help clean check-vulnerabilities
```
jonathaningram: Did you need to actually add the empty target to this Makefile?

utils/Makefile
```diff
@@ -27,6 +27,9 @@ build:
 tests:
 	$(RUNNER) python setup.py test
 
+check-vulnerabilities:
```
jonathaningram: Add to .PHONY?

Makefile
```diff
@@ -59,7 +59,7 @@ cf-login-prod:
 	  cf-login
 
 DIRS        = api admin utils status
-TARGETS     = setup setup-dev build clean deploy deploy-dev test
+TARGETS     = setup setup-dev build check-vulnerabilities clean deploy deploy-dev test
```
jonathaningram: If we run `make check-vulnerabilities` from the project root, does this succeed? Per my comment below, I wonder if it will fail because the status project is missing an empty target. I could be wrong though.

.circleci/config.yml
```diff
@@ -123,6 +123,17 @@ jobs:
...
+  check-vulnerabilities:
+    <<: *python_node_postgres_images
+    steps:
+      - *attach_workspace
+      - run: make admin.setup-dev
+      - run: make admin.check-vulnerabilities
+      - run: make api.setup-dev
+      - run: make api.check-vulnerabilities
+      - run: make utils.setup-dev
+      - run: make utils.check-vulnerabilities
```
jonathaningram: I guess with this section, would it be better do it like this:

```
- run: make setup-dev
- run: make check-vulnerabilities
```

?

Meaning that if we add checks to the status project (and others we add), this will run those too.

Makefile
dagoof: yeah it will fail on status.

we could extend this makefile target to include checking for out of date dependencies like I did when I was updating everything.


There's a ruby tool that I used called https://github.com/rubysec/bundler-audit

`pipenv check` and `npm audit` do the same for python and node projects, respectively.

dagoof: Not sure that i would want to fail CI on out of date deps though.

jonathaningram: Yeah I don't know about failing CI for that either, but good to know.

status/Makefile
andrewoh531: Nope, it's been removed

utils/Makefile
andrewoh531: Thanks

Makefile
andrewoh531: Is this the prefered way to do things? I was planning to do it through this and have empty targets in `status/` but went the other direction and explicitly call `check-vulnerabilities` for each of the 3 directories instead; so using that method I can remove `check-vulnerabilities` from `TARGETS`

.circleci/config.yml
andrewoh531: Changed

.circleci/config.yml
```diff
@@ -123,6 +123,15 @@ jobs:
       - run: echo 're-enable tests here'
       - run: echo 'make utils.test'
 
+  check-vulnerabilities:
+    <<: *python_node_postgres_images
+    steps:
+      - *attach_workspace
+      - run: make admin.setup-dev
+      - run: make api.setup-dev
+      - run: make utils.setup-dev
```
jonathaningram: Is this not just `- run: make setup-dev`?

andrewoh531: `make setup-dev` will also run for the `status` directory which isn't needed for now

jonathaningram: OK no worries.

utils/notifications_utils/template.py
```diff
@@ -38,11 +38,12 @@
...
-    path.join(
-        path.dirname(path.abspath(__file__)),
-        'jinja_templates',
-    )
+template_env = Environment(autoescape=select_autoescape(['html', 'htm', 'xml']),
+    loader=FileSystemLoader(
+        path.join(
+            path.dirname(path.abspath(__file__)),
+            'jinja_templates',
+        )
```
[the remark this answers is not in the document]
andrewoh531: I've created a `feat-bandit` branch for testing: https://github.com/govau/notify/pull/87.

Closing this in favour of the above branch.

andrewoh531:
Superseded by: https://github.com/govau/notify/pull/87

PR #85, opened 2018-09-27
feat(deploy): use ZDT for deploy branches too

jonathaningram:
This is what prod uses. By not using this for feature branches,
we can break master because changes in manifest.yml, for example
are not tested.


jonathaningram:
Relevant for why the build fails: https://github.com/cloudfoundry/cli/issues/1399

PR #86, opened 2018-10-01
Adding CREATE_ADMIN_USER=false to .env.sample

andrewoh531:
Will be needed locally when migration scripts are run (e.g. running tests)

### DIFF
--- a/api/.env.sample
+++ b/api/.env.sample
@@ -27,4 +27,6 @@ TWILIO_ACCOUNT_SID=
 TWILIO_AUTH_TOKEN=
 TWILIO_FROM_NUMBER=
 
+CREATE_ADMIN_USER=false
+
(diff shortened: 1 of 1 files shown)


api/.env.sample
```diff
@@ -27,4 +27,6 @@ TWILIO_ACCOUNT_SID=
 TWILIO_AUTH_TOKEN=
 TWILIO_FROM_NUMBER=
 
+CREATE_ADMIN_USER=false
```
jonathaningram: Why is this needed locally? Won't the migration just gracefully exit if the env var is not set?

https://github.com/govau/notify/blob/3e886de5e8943418602ef751a4c2dc286bc5a8ff/api/migrations/versions/0200_create_admin_user.py#L24

andrewoh531: Actually you're right...too early in the week

PR #87, opened 2018-10-02
Integrate python Bandit with CI

andrewoh531:


utils/notifications_utils/template.py
```diff
@@ -38,11 +38,12 @@
...
-    path.join(
-        path.dirname(path.abspath(__file__)),
-        'jinja_templates',
-    )
+template_env = Environment(autoescape=select_autoescape(['html', 'htm', 'xml']),
+    loader=FileSystemLoader(
+        path.join(
+            path.dirname(path.abspath(__file__)),
+            'jinja_templates',
+        )
```
jonathaningram: LGTM. Can't see any obvious issues but I didn't test every page or function.

andrewoh531: We'll test in production 😄

PR #89, opened 2018-10-02
Feat doc spike

dagoof:
hey. it's close enough; let's get this merged so we can work on it.

I'll be squashing these commits


docs/LICENSE
```diff
@@ -0,0 +1,22 @@
+The MIT License (MIT)
```
jonathaningram: Double checking, but adding the Gatsby license was intentional yes? Because:

> The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.

docs/package.json
```diff
@@ -0,0 +1,41 @@
...
+    "develop": "gatsby develop",
+    "format": "prettier --write '**/*.js'",
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "devDependencies": {
+    "prettier": "^1.14.2"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/gatsbyjs/gatsby-starter-default"
```
jonathaningram: Need to remove this, but can do later.

docs/src/components/base.css
```diff
@@ -0,0 +1,226 @@
+html {
```
jonathaningram: Out of interest where is the base/core CSS from?

docs/LICENSE
dagoof: totally not intentional

docs/src/components/base.css
dagoof: gatsby included "layout.css". it had a bunch of stuff that was ugly so I stripped it back to create this monster. we need to replace the whole thing with a design system

PR #91, opened 2018-10-08
Logo CDN

jonathaningram:


jonathaningram:
@dagoof @andrewoh531 can be tested once staging has the env var set and the apps are restaged.

admin/app/config.py
```diff
@@ -15,6 +15,8 @@ class Config(object):
     # if we're not on cloudfoundry, we can get to this app from localhost. but on cloudfoundry its different
     ADMIN_BASE_URL = os.environ.get('ADMIN_BASE_URL', 'http://localhost:6012')
 
+    CDN_BASE_URL = os.environ.get('CDN_BASE_URL', 'https://notify-static-logos-staging.static.cld.gov.au')
```
andrewoh531: Is `CDN_BASE_URL` for prod?

jonathaningram: Are you thinking that dev should just use a localhost CDN? If so, I was thinking that dev should be as close to prod as possible. Thoughts?

andrewoh531: I'm ok with dev (non-prod) pointing to `https://notify-static-logos-staging.static.cld.gov.au`. I was wondering where we're setting `CDN_BASE_URL` to `https://static-logos.notify.gov.au` (looks like within CF UPS)

jonathaningram: Yeah same with how we set the S3 buckets, I wanted to keep the prod config in UPS and definitely not default to it (especially for a CDN that would serve an image for a long time). I was deciding if I should even have a default here, or if non-prod should also specify their URL explicitly, but just decided to put in the default in config.py. I guess at some point we should clean up the inconsistency that exists between the defaults and envs.

admin/app/main/s3_client.py
```diff
@@ -39,9 +39,11 @@ def get_object(self, filename):
     def delete_object(self, filename):
         return self.get_object(filename).delete()
 
-    def copy_object(self, old_name, new_name):
+    def copy_object(self, old_name, new_name, acl=None):
```
dagoof: depends how explicit you want this to be. Personally I don't mind forwarding arguments to copy_from. you would do this with something like

```
def copy_object(self, old_name, new_name, **kwargs):
    ...copy_from(..., **kwargs)
```

PR #92, opened 2018-10-08
feat(cf): add makefile target to diff UPS vars

dagoof:


andrewoh531:
Is this the final solution to our secrets management or an intermediate tool?

jonathaningram:
> Is this the final solution to our secrets management or an intermediate tool?

@dagoof? I see this as intermediate.

Makefile
```diff
@@ -80,6 +82,13 @@ $(SVC_APPLIED): apply-service-%: ci/ups/$(CLD_HOST)/%.json
 $(SVC_CREATED): create-service-%:
 	$(MAKE) apply-service-$* APPLY_ACTION=create
 
+$(SVC_DIFFED): SHELL = /bin/bash
+$(SVC_DIFFED): diff-service-%:
+	@diff \
```
jonathaningram: @dagoof could we use `sdiff` or `diff -s`? More useful output IMO.

dagoof: tried sdiff, why is it more useful here? instead of listing the changes mine is now echoing out both identical files

PR #93, opened 2018-10-09
Fix downgrade script

andrewoh531:


api/migrations/versions/0199_change_provider_details.py
```diff
@@ -57,7 +57,7 @@ def make_provider_details_downgrade_sql(identifier):
             ORDER BY version desc
             LIMIT 1
         ) h
-        WHERE identifier = '{0}'
+        WHERE provider_details.identifier = '{0}'
```
jonathaningram: @andrewoh531 is this required?

api/migrations/versions/0199_change_provider_details.py
```diff
@@ -35,7 +35,7 @@ def make_provider_details_history_upgrade_sql(identifier):
 def make_provider_details_history_downgrade_sql(identifier):
     return """
         DELETE FROM provider_details_history
-        WHERE identifier = {0} AND version = (SELECT version from provider_details WHERE identifier = '{0}')
+        WHERE identifier = '{0}' AND version = (SELECT version from provider_details WHERE identifier = '{0}')
```
jonathaningram: @andrewoh531 thanks.

api/migrations/versions/0199_change_provider_details.py
dagoof: sql complaints that it is ambiguous otherwise

jonathaningram: Cool, thanks for fixing my rubbish SQL.

andrewoh531: Yeah it was complaining of ambiguous column name (line number in the logs doesn't match up to the file):
```
sqlalchemy.exc.ProgrammingError: (psycopg2.ProgrammingError) column reference "identifier" is ambiguous
LINE 19:         WHERE identifier = 'firetext'
```

PR #95, opened 2018-10-10
Adding code samples component

andrewoh531:


docs/src/components/code-examples.js
```diff
@@ -1,9 +1,150 @@
...
+                customStyle={customStyle}
+                language={language}
+                children={value}
+              />
+            ),
+          }}
+        />
+      )
+    } else {
+      return (
```
jonathaningram: Just return early, no else.

docs/src/components/code-examples.js
```diff
@@ -1,9 +1,150 @@
...
+                  {s.name}
+                </NavLink>
+              </NavItem>
+            ))}
+          </Ul>
+        </TabsNav>
+
+        <div>
+          {codeSnippets.map((s, i) => (
+            <TabContent key={i} active={this.state.activeTab === i}>
```
jonathaningram: Does the `node` for this snippet have an identifier? `i` is not reliable as a value for `key` because a different set of `codeSnippets` of the same length will have the same values for `id` but different nodes. Best to use an identifier, or the closest possible identifier of the iterated item.

docs/src/components/code-examples.js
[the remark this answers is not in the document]
andrewoh531: Done

docs/src/components/code-examples.js
```diff
@@ -1,9 +1,150 @@
...
+                  active={this.state.activeTab === i}
+                >
+                  {s.name}
+                </NavLink>
+              </NavItem>
+            ))}
+          </Ul>
+        </TabsNav>
+
+        <div>
```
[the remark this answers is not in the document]
andrewoh531: Done

docs/src/pages/installation/Python.py
```diff
@@ -0,0 +1 @@
+pip install --upgrade govau-platforms-notify
```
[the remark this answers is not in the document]
andrewoh531: Changed

docs/src/pages/sending-text-messages.mdx
```diff
@@ -4,11 +4,12 @@ import Examples from '../components/code-examples'
 
 export default Layout
 
+
 ## Sending a single text message
 
 The following example shows how to send a single text message.
 
-<Examples subject="sending-text-messages" />
+<Examples codePath="sending-texts"/>
```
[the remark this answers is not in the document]
andrewoh531: I'll change it to `path`

docs/src/pages/sending-emails/Java.java
```diff
@@ -0,0 +1,8 @@
+HashMap<String, String> personalisation = new HashMap<String, String>();
+personalisation.put("user_name", "Lisa");
+
+SendEmailResponse response = client.sendEmail(
+    "f33517ff-2a88-4f6e-b855-xxxxxxxxxxxx",
+    "lisa@example.com",
+    personalisation,
+    "your-unique-reference-id");
```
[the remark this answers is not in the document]
andrewoh531: @jonathaningram Not sure I follow

docs/src/pages/installation.mdx
```diff
@@ -8,7 +8,7 @@ export default Layout
 
 Install the client libraries in your choice of programming language
 
-<Examples subject="installation" />
+<Examples codePath="installation"/>
```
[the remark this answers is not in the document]
andrewoh531: Doesn't look like this has been released yet. To be [released](https://github.com/prettier/prettier/pull/4975) in `1.15` which isn't out yet unless we want to use the dev version

docs/src/components/code-examples.js
andrewoh531: If I'm understanding you correctly are you saying that key should be globally unique? It looks like it only needs to be [unique amongst its siblings](https://reactjs.org/docs/lists-and-keys.html#keys): `Keys used within arrays should be unique among their siblings. However they don’t need to be globally unique. We can use the same keys when we produce two different arrays`

docs/src/components/code-examples.js
```diff
@@ -1,9 +1,150 @@
...
+  display: ${props => (props.active ? 'visible' : 'none')};
+`
+
+export default class CodeExamples extends React.Component {
+  constructor(props) {
+    super(props)
+    this.state = { activeTab: 0 }
+  }
+
+  onClick = index => {
```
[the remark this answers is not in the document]
andrewoh531: Changed

docs/src/components/code-examples.js
```diff
@@ -1,9 +1,150 @@
...
+`
+
+const TabContent = styled.div`
+  display: ${props => (props.active ? 'visible' : 'none')};
+`
+
+export default class CodeExamples extends React.Component {
+  constructor(props) {
+    super(props)
+    this.state = { activeTab: 0 }
```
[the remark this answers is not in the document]
andrewoh531: Changed

docs/src/pages/sending-emails/Java.java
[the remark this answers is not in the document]
andrewoh531: Thanks

docs/src/components/code-examples.js
```diff
@@ -1,9 +1,150 @@
...
+  padding: 0;
+  display: inline-block;
+`
+
+const NavLink = styled.label`
+  ${linkCSS};
+`
+
+const TabContent = styled.div`
+  display: ${props => (props.active ? 'visible' : 'none')};
```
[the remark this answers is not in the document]
andrewoh531: Sorry, can you elaborate?

docs/src/components/code-examples.js
```diff
@@ -1,9 +1,150 @@
...
+  constructor(props) {
+    super(props)
+    this.state = { activeTab: 0 }
+  }
+
+  onClick = index => {
+    this.setState({ activeTab: index })
+  }
+
+  filterTransformSortCodeSnippets = codeSnippets =>
```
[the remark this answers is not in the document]
andrewoh531: Yeah ok

docs/loadershim.js
```diff
@@ -0,0 +1,6 @@
+global.___loader = {
+  enqueue: jest.fn(),
+}
+
+const gatsby = jest.requireActual('gatsby')
+module.exports = { ...gatsby, graphql: jest.fn(), Link: 'Link' }
```
jonathaningram: What's the `Link` for?

docs/package.json
```diff
@@ -22,9 +24,44 @@
...
+    "@babel/plugin-proposal-optional-chaining": "^7.0.0",
+    "@babel/preset-env": "^7.1.0",
+    "@babel/preset-react": "^7.0.0",
+    "babel-core": "^7.0.0-bridge.0",
+    "babel-jest": "^23.6.0",
+    "enzyme": "^3.7.0",
+    "identity-obj-proxy": "^3.0.0",
+    "jest": "^23.6.0",
+    "prettier": "^1.14.2",
+    "react-test-renderer": "^16.5.2"
```
jonathaningram: I think we should just move everything to `dependencies`. It doesn't have to block this PR. But there's no real need to separate them. It's often not obvious where a dep should go anyway, so all in `dependencies` is fine.

docs/src/components/code-examples.test.js
```diff
@@ -0,0 +1,25 @@
...
+  ${undefined} | ${'sending-emails'} | ${0}           | ${'returns empty array when data is undefined'}
+`(
+  'filterTransformSortCodeSnippets $testDescription',
+  ({ data, path, expectedLength }) => {
+    expect(
+      filterTransformSortCodeSnippets({
+        codeSnippets: mockGraphQLResponse.allCodeSamples.edges,
+        path: 'sending-emails',
+      })
+    ).toHaveLength(3)
```
jonathaningram: `expectedLength`?

docs/src/components/code-examples.test.js
```diff
@@ -0,0 +1,25 @@
...
+
+const mockData = mockGraphQLResponse.allCodeSamples.edges
+
+test.each`
+  data         | path                | expectedLength | testDescription
+  ${mockData}  | ${'sending-emails'} | ${3}           | ${'filters correctly'}
+  ${mockData}  | ${'unknown-path'}   | ${0}           | ${'returns empty array no matches found'}
+  ${[]}        | ${'sending-emails'} | ${0}           | ${'returns empty array when data is empty'}
+  ${null}      | ${'sending-emails'} | ${0}           | ${'returns empty array when data is null'}
+  ${undefined} | ${'sending-emails'} | ${0}           | ${'returns empty array when data is undefined'}
```
jonathaningram: Regarding `null` and `undefined` as passed to this func, do you know what graphql actually passes if there are no results? Because if it's undefined you can change the actual signature of the function to just:

```js
export const filterTransformSortCodeSnippets = ({ codeSnippets = [], path }) => {
```
And remove the check at the start. If graphql passes `null` then unfortunately we need to keep it in there.

docs/src/components/code-examples.test.fixture.js
```diff
@@ -0,0 +1,97 @@
+const mockGraphQLResponse = {
+  allCodeSamples: {
+    edges: [
+      {
+        node: {
+          content:
+            'import au.gov.notify.NotifyClient;\n\nNotifyClient client = new NotifyClient(apiKey);\n',
```
jonathaningram: I like that this is real content, but wonder if it's just worth keeping it more concise for the test fixtures? E.g. even the `Java.md` fixture really only needs to have content like `hi I am _markdown_` if you were going to use this fixture in the future to check that it renders the correct HTML.

docs/package.json
dagoof: I don't really care but you could make a case for things that require the build to go in deps while things like test, prettier, go into devdeps

docs/src/components/code-examples.js
```diff
@@ -1,9 +1,123 @@
...
+    const codeSnippets = filterTransformSortCodeSnippets({
+      codeSnippets: this.props.data.allCodeSamples.edges,
+      path: this.props.path,
+    })
+    return (
+      <Tabs>
+        <TabsNav>
+          <Ul>
+            {codeSnippets.map((s, i) => (
+              <NavItem key={i}>
```
jonathaningram: @andrewoh531 can't find your previous comment regarding `key`. But, the value has to be unique across items that are `map`ped. E.g. if you do this:

```jsx
const arr = ["a", "b", "c"];
arr.map((v, i) => <span key={i}>{v}</span>)
```

Then if the values in array change, it won't necessarily re-render, e.g.:

```jsx
const arr = ["c", "a", "b"]
```

Because the `i`ndex is still `0`, `1` and `2`.

See https://reactjs.org/docs/lists-and-keys.html#keys

> The best way to pick a key is to use a string that uniquely identifies a list item among its siblings. Most often you would use IDs from your data as keys

And

> We don’t recommend using indexes for keys if the order of items may change.

docs/src/components/code-examples.js
```diff
@@ -1,9 +1,123 @@
...
+                >
+                  {s.name}
+                </NavLink>
+              </NavItem>
+            ))}
+          </Ul>
+        </TabsNav>
+
+        {codeSnippets.map((s, i) => (
+          <TabContent key={i} active={this.state.activeTab === i}>
```
jonathaningram: Same here regarding `key`.

docs/src/components/code-examples.js
jonathaningram: If you're unsure of a good value for the "id", let's chat about it, but I guess you're just trying to find out what is unique about the code snippet and use that.

docs/src/components/code-examples.js
```diff
@@ -1,9 +1,123 @@
...
+      codeSnippets: this.props.data.allCodeSamples.edges,
+      path: this.props.path,
+    })
+    return (
+      <Tabs>
+        <TabsNav>
+          <Ul>
+            {codeSnippets.map((s, i) => (
+              <NavItem key={i}>
+                <NavLink
```
jonathaningram: Pretty sure we want this to be an `a`nchor element rather than a label. E.g. see https://designsystem.gov.au/components/buttons/ and tab through the page - aside from what they do, I'd expect to be able to tab through the page and when I get to "Python" or "Go", for example, to hit 'enter' and have it change to that tab.

docs/loadershim.js
andrewoh531: Probably not needed

docs/src/components/code-examples.test.js
andrewoh531: lol yup

docs/package.json
andrewoh531: I'm not fussed either but adding them to `dependencies` will increase the `build-docs` step which is a pre-requisite for the deploy.

docs/src/components/code-examples.js
andrewoh531: Thanks for the info

docs/src/components/code-examples.test.fixture.js
andrewoh531: I can see your point, but I also see an advantage in keeping it close to real thing. Taking your example if we need to add more tests we're able to use this fixture without changing it and tests that are already using it don't need to be changed as a result.

docs/package.json
jonathaningram: Am I missing something, because in circle, `npm install` is run which will install all deps from package.json. Also, I would see that the way to do it is to:

- npm install
- cache node_modules
- restore cached node_modules and test
- restore cached node_modules and build

I've just found it not obvious where things go so default to just `dependencies`. Makes me have to make less decisions.

docs/src/components/code-examples.test.fixture.js
jonathaningram: No worries. Let's leave it as you have it. If we find we're actually copy/pasting the real content, then it means we're probably wanting to test the real thing, so we'll see if we can import the real thing as a fixture.

Out of interest would it be easy to write `docs/src/components/code-examples.test.fixture.js` like:

```
const content = require('../../file.md')
{
        node: {
          content,
       }
}
```

Would that work?

docs/package.json
andrewoh531: Sorry you're right, I assumed we had `--production` flag for installing. This is going for awhile. I'll move them into `dependencies`

docs/src/components/code-examples.test.fixture.js
andrewoh531: I think it would, but to clarify I didn't manually copy the contents from each `md` file to here; I ran the graphQL query and just pasted the response from there.

docs/src/components/code-examples.js
```diff
@@ -1,9 +1,125 @@
...
+}
+
+export default ({ path }) => (
+  <StaticQuery
+    query={graphql`
+      query {
+        allCodeSamples {
+          edges {
+            node {
+              content
```
jonathaningram: Could pull in `id`?

docs/src/components/code-examples.js
```diff
@@ -1,9 +1,125 @@
...
+    const codeSnippets = filterTransformSortCodeSnippets({
+      codeSnippets: this.props.data.allCodeSamples.edges,
+      path: this.props.path,
+    })
+    return (
+      <Tabs>
+        <TabsNav>
+          <Ul>
+            {codeSnippets.map((s, i) => (
+              <NavItem key={s.relativePath}>
```
jonathaningram: Then use `s.id`?

docs/src/components/code-examples.js
```diff
@@ -1,9 +1,125 @@
...
+                >
+                  {s.name}
+                </NavLink>
+              </NavItem>
+            ))}
+          </Ul>
+        </TabsNav>
+
+        {codeSnippets.map((s, i) => (
+          <TabContent key={s.relativePath} active={this.state.activeTab === i}>
```
jonathaningram: And `s.id`?

docs/src/components/code-examples.js
```diff
@@ -1,9 +1,125 @@
...
+const TabContent = styled.div`
+  display: ${props => (props.active ? 'block' : 'none')};
+`
+
+export const filterTransformSortCodeSnippets = ({ codeSnippets, path }) => {
+  if (!codeSnippets) {
+    return []
+  }
+
+  return codeSnippets
```
jonathaningram: Minor, but could just do `return (codeSnippets || [])` and remove the conditional at the start of the function.

PR #99, opened 2018-10-12
feat(fullstory): add fullstory js to deployed apps

dagoof:


dagoof:
going to close this until we've blocked sensitive elements in fullstory.

PR #105, opened 2018-10-15
Styling the footer

andrewoh531:


docs/src/components/footer.js
```diff
@@ -1,55 +1,106 @@
...
   margin-bottom: 0;
+  list-style-type: none;
 `
 
+const UnorderedList = ({ list }) => (
+  <Box width={[1, 1 / 4]}>
+    <Ul>
+      {list.map(i => (
+        <li key={i.text}>
+          <External blank href={i.target}>
```
jonathaningram: I think we can leave `blank` off - user can open in new window if they want.

docs/src/components/footer.js
```diff
@@ -1,55 +1,106 @@
...
+            {i.text}
+          </External>
+        </li>
+      ))}
+    </Ul>
+  </Box>
+)
+
+const firstColumn = [
+  { text: 'Support', target: 'https://notify.gov.au/support' },
```
jonathaningram: `target` -> `href`

docs/src/components/footer.js
andrewoh531: Removed

docs/src/components/footer.js
andrewoh531: Changed

docs/src/components/footer.js
```diff
@@ -1,55 +1,104 @@
...
+
+  justify-content: space-between;
   align-items: start;
+  border-bottom: 1px solid #bfc1c3;
+  padding-bottom: 4rem;
 
-  & > * + * {
-    margin-left: 4rem;
-  }
+  ${space};
```
jonathaningram: `<Flex/>` has space props already so you don't need this, will share code sample offline.

PR #106, opened 2018-10-16
Fix docs deploy

andrewoh531:


.circleci/config.yml
```diff
@@ -115,6 +115,8 @@ jobs:
           root: .
           paths:
             - docs
+            - docs/node_modules
```
jonathaningram: Are these required? Doesn't just `docs` cover both of these?

andrewoh531: It shouldn't be required. It was failing before with `sh: 1: gatsby: not found` without it and the `node_modules` wasn't being persisted. It looks to be working now.

PR #108, opened 2018-10-18
Adding skip nav functionality

andrewoh531:


docs/src/components/skip-nav.js
```diff
@@ -0,0 +1,31 @@
...
+    left: 10px;
+    background-color: ${props => props.theme.highlight};
+    z-index: 1;
+    width: auto;
+    height: auto;
+    clip: auto;
+  }
+`
+
+export const SkipNavLink = props =>
```
jonathaningram: Just export the component above and let callers use it.

PR #112, opened 2018-10-23
Styling header

andrewoh531:
Can view changes here: https://notify-docs-f-header.apps.y.cld.gov.au


docs/src/components/header.js
```diff
@@ -1,70 +1,223 @@
...
 
-const Banner = styled.h1`
-  display: inline-block;
-  font-weight: normal;
-  font-size: 1.8em;
-  margin: 0 2rem;
+const BannerLink = styled.a`
+  text-decoration: none;
+
+  & :hover {
```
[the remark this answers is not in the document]
andrewoh531: Do you mean from a code style point of view? Yeah

docs/src/components/header.js
```diff
@@ -1,70 +1,223 @@
...
+  text-decoration: none;
+  color: ${props => (props.active ? '#45c2f0' : '#fff')};
+
+  ${util.createMediaQuery(util.defaultBreakpoints[0])} {
+    color: white;
+  }
 `
 
-export default ({ siteTitle }) => (
+const MyNavItem = ({ children, href, active, ...props }) => (
```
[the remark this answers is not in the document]
andrewoh531: lol sorry that's badly named. I've changed it to BoxedListItem for now but open to suggestions

docs/src/components/header.js
```diff
@@ -1,70 +1,223 @@
...
-        </Heading>
+        <Flex justifyContent="space-between">
+          <BannerLink to="/">
+            <Flex justifyContent="flex-start" p={'3rem 0'} alignItems="center">
+              <Logo src={dta} />
+              <HeaderName>Notify</HeaderName>
+              <Badge>Alpha</Badge>
+            </Flex>
+          </BannerLink>
+          <HamburgerIcon href="#nav" aria-label="Open Navigation">
```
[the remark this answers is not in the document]
andrewoh531: Changed

docs/src/components/header.js
```diff
@@ -1,70 +1,223 @@
...
+            <FontAwesomeIcon icon="bars" />
+          </HamburgerIcon>
+        </Flex>
+
+        <Nav id="nav">
+          <CloseIcon href="#" aria-label="Close Navigation">
```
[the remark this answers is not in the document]
andrewoh531: Changed

docs/package.json
```diff
@@ -5,12 +5,16 @@
...
+    "@fortawesome/fontawesome-svg-core": "^1.2.6",
+    "@fortawesome/free-solid-svg-icons": "^5.4.1",
+    "@fortawesome/react-fontawesome": "^0.1.3",
     "@mdx-js/mdx": "^0.15.4",
     "@mdx-js/tag": "^0.15.0",
     "@rebass/grid": "^6.0.0-4",
     "babel-core": "^7.0.0-bridge.0",
     "babel-jest": "^23.6.0",
     "babel-plugin-styled-components": "^1.7.1",
+    "core-js": "^2.5.7",
```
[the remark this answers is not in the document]
andrewoh531: It seems to be related to [some changes in Babel's use of `core-js`](https://github.com/gatsbyjs/gatsby/issues/7862#issuecomment-423855759). The workaround that was floating around in there was to add `core-js` as a depedency.

docs/src/components/header.js
```diff
@@ -1,70 +1,234 @@
...
+            <MenuText>Menu</MenuText>
+            <MenuIcon />
+          </OpenMenuLink>
+        </Flex>
+
+        <Nav id="nav">
+          <CloseLink href="#" aria-label="Close navigation">
+            <CloseIcon />
+          </CloseLink>
+          <StyledUnsortedList>
```
jonathaningram: See below. This is a `ul` but the children are not `li`s.

image

docs/src/components/header.js
```diff
@@ -1,70 +1,234 @@
...
+            <Flex flexWrap="wrap" m={[0, '1rem 0 1rem -1.5rem']}>
+              <Box width={[1, 'auto']}>
```
jonathaningram: Can you look into the `width` attribute being forwarded on to the DOM? It shouldn't be—I've come across this before but can't recall why it does it. Grid should just use it to determine the CSS but shouldn't be forwarding it to the DOM element.

docs/src/components/header.js
```diff
@@ -1,70 +1,234 @@
...
+
+  ${util.createMediaQuery(util.defaultBreakpoints[0])} {
+    color: white;
+  }
 `
 
-export default ({ siteTitle }) => (
+const BoxedListItem = ({ children, href, active, ...props }) => (
+  <Box mb={['0.7rem', '0.1rem']} width={[1, 'auto']} {...props}>
+    <StyledListItem active={active}>
```
jonathaningram: This is also an `li` not inside a `ul` parent.

docs/src/components/header.js
```diff
@@ -1,70 +1,234 @@
...
+
+        <Nav id="nav">
+          <CloseLink href="#" aria-label="Close navigation">
+            <CloseIcon />
+          </CloseLink>
+          <StyledUnsortedList>
+            <Flex flexWrap="wrap" m={[0, '1rem 0 1rem -1.5rem']}>
+              <Box width={[1, 'auto']}>
+                <Flex flexWrap="wrap">
+                  <BoxedListItem href={`${notifyBaseUrl}/support`}>
```
jonathaningram: Do we need the `BoxedListItem` wrapper? Can we just have the basic component structure of:

```jsx
<Links>
  <LinkItem>
    <ExternalLink ...>Foo</ExternalLink>
    <ExternalLink ...>Foo</ExternalLink>
    <ExternalLink ...>Foo</ExternalLink>
<Links>
```

And each of those components can be a styled component themselves (doesn't necessarily have to be named `StyledX`). I think we should be able to apply the necessary CSS to the `li` to get it to flex how we want it.

Also this lets the children be anything later not just `ExternalLink` which is all we can have now, e.g.:

```jsx
<Links>
  <LinkItem>
    <ExternalLink ...>Foo</ExternalLink>
    <ExternalLink ...>Foo</ExternalLink>
    <Link to="/react-router-link">Foo</Link>
<Links>
```

docs/src/components/header.js
andrewoh531: Changed

docs/src/components/header.js
andrewoh531: Changed

PR #115, opened 2018-10-25
Fix Notify's mobile view

andrewoh531:
Changes can be viewed here - https://notify-f-mobile-ui.apps.y.cld.gov.au/features


admin/app/templates/govau_template.html
```diff
@@ -53,6 +53,9 @@
               <span>{{ global_header_text|default('GOV.AU') }}</span>
               <span class="header-badge">alpha</span>
             </a>
+            <a href="#proposition-menu" id="open-navigation">Menu 
+              <img src="{{ asset_path }}images/navigation-open.svg" alt="Open navigation"/>
+            </a>
           </div>
           {% block inside_header %}{% endblock %}
         </div>
```
jonathaningram: Not sure if you had planned to do it in this PR but can we have the home page look like the following?

image

admin/app/templates/admin_template.html
```diff
@@ -43,14 +43,16 @@
 {% block proposition_header %}
   <div class="header-proposition">
     <div class="content">
-      <a href="#proposition-links" class="js-header-toggle menu">Menu</a>
       <nav id="proposition-menu">
+        <a href="#" id="close-navigation">
```
jonathaningram: ARIA?

admin/app/templates/govau_template.html
```diff
@@ -53,6 +53,9 @@
               <span>{{ global_header_text|default('GOV.AU') }}</span>
               <span class="header-badge">alpha</span>
             </a>
+            <a href="#proposition-menu" id="open-navigation">Menu
```
jonathaningram: ARIA?

admin/app/templates/govau_template.html
andrewoh531: Made some changes in this area. Let me know what you think @jonathaningram

jonathaningram:
@andrewoh531 see:

image

Not sure if the changes in this PR caused the loss of left and right space on the above, but either way, can you take a look at fixing it as part of this PR?

jonathaningram:
@dagoof 

> is this meant to be comprehensive or just for unauthenticated pages? once you start logging in it gets a bit weird

We have another ticket for auth pages.

PR #119, opened 2018-10-30
Adding support for DOCS_BASE_URL to set docs base url

andrewoh531:


admin/app/config.py
```diff
@@ -16,6 +16,7 @@ class Config(object):
     ADMIN_BASE_URL = os.environ.get('ADMIN_BASE_URL', 'http://localhost:6012')
 
     CDN_BASE_URL = os.environ.get('CDN_BASE_URL', 'https://notify-static-logos-staging.static.cld.gov.au')
+    DOCS_BASE_URL = os.environ.get('DOCS_BASE_URL', 'https://docs.notify.gov.au')
```
jonathaningram: Cool, you can now use the env var in templates? i.e. change hard-coded cases to use this config.

andrewoh531: I'll do it in a separate branch

PR #120, opened 2018-10-30
Retrieve docs base url via app config

andrewoh531:
Can view the link from this page: https://notify-f-docs-links.apps.y.cld.gov.au/documentation


admin/manifest-dev.yml
```diff
@@ -19,4 +19,5 @@ applications:
     env:
       <<: *defaults_env
       ADMIN_BASE_URL: https://notify-((stg)).apps.y.cld.gov.au
+      DOCS_BASE_URL: https://notify-docs-((stg)).apps.y.cld.gov.au/
```
jonathaningram: I think you need to remove the trailing slash here: 

image

PR #126, opened 2018-11-05
Fixing left alignment on authenticated pages

andrewoh531:


jonathaningram:
@andrewoh531

Some issues.

First showing unauth cookie bar not aligned. Second showing footer indented from horizontal rule:

image
image

Third showing auth not right either:

image

You might like to revert your local back to the previous change (the mobile UI changes) and check this again. I say this because it looks like that PR and this PR have resulted in the whole UI main body shrinking in size.

PR #127, opened 2018-11-05
feat(celery): add celery beat for scheduled tasks

jonathaningram:
We're not currently running any scheduled tasks. This change adds the
celery beat application to our build.


api/manifest-dev.yml
```diff
@@ -29,3 +29,9 @@ applications:
     health-check-type: process
     no-route: true
     command: make run-celery
+  - name: notify-celery-beat-((stg))
+    <<: *defaults
+    memory: 256M
+    health-check-type: process
+    no-route: true
+    command: make run-celery
```
andrewoh531: `command: make run-celery-beat`?

jonathaningram:
@andrewoh531 can you please review again? We should both do actual tests on the build for this one. Also before I approve and merge to master we should test stg.

api/app/celery/research_mode_tasks.py
```diff
@@ -11,34 +11,26 @@
 from app.config import QueueNames
 from app.celery.process_ses_receipts_tasks import process_ses_results
 
-temp_fail = "7700900003"
-perm_fail = "7700900002"
-delivered = "7700900001"
+temp_fail = "409000003"
```
andrewoh531: Do you know what research mode is? I can see it's one of the modes a service can be placed under but there isn't documentation around it I could find.

jonathaningram: `research_mode` is a property on a service that can be enabled by platform admin.

PR #131, opened 2018-11-08
fix(invite): remove leading white space in tmpl

jonathaningram:
The leading white space causes the paragraph to be formatted as
preformatted text when converted from Markdown to HTML.

### DIFF
--- a/api/migrations/versions/0205_fix_email_formatting.py
+++ b/api/migrations/versions/0205_fix_email_formatting.py
@@ -1,0 +1,62 @@
+"""
+
+Revision ID: 0205
+Revises: 0204
+Create Date: 2018-11-12 15:32:20.317286
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+revision = '0205'
+down_revision = '0204'
+
+
+templates = [{
+    'id': "4f46df42-f795-4cc4-83bb-65ca312f49cc",
+    'new_content': """((user_name)) has invited you to collaborate on ((service_name)) using Notify.
+
+Notify makes it easy to keep people updated by helping you send text messages and emails.
+
+Use the following link to create a Notify account:
+
+((url))
+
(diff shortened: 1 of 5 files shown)


jonathaningram:
@dagoof @andrewoh531 as you both know, there's the weird issue of spacing being added to this email. This PR does not actually fix that but it just fixes the wrong preformatted text. Will add the weird space issue as a new ticket. That's a less intrusive bug than preformatted text so would rather get this fixed first.

api/migrations/versions/0205_fix_email_formatting.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+revision = '0205'
+down_revision = '0204'
+
+
+templates = [{
+    'id': "4f46df42-f795-4cc4-83bb-65ca312f49cc",
+    'new_name': "Notify invitation email",
+    'new_subject': "((user_name)) has invited you to collaborate on ((service_name)) on Notify.gov.au",
```
dagoof: I'm no copy expert. but this is the same length and reads worse than our content

collaborate ON x ON notify.gov

utils/.flake8
```diff
@@ -3,4 +3,4 @@
 # W503: line break before binary operator
 exclude = venv*,__pycache__,cache
 ignore = W503,E501
-max-complexity = 8
+max-complexity = 14
```
dagoof: 💯

api/migrations/versions/0205_fix_email_formatting.py
jonathaningram: I didn't change it though. Only fixed the preformatted text issue.

jonathaningram: Good pickup. Well, kind of. I don't actually update the subject in this migration so will remove the keys from the dict.

PR #132, opened 2018-11-09
Fix mobile views

andrewoh531:


jonathaningram:
@andrewoh531 we've lost some padding on the service heading:

Compare space under 'Test' in your changes:

image

With live:

image

PR #136, opened 2018-11-12
feat(docs-language): lang selector on example

dagoof:
remove language selector from sidenav

swap language selector dropdown for simpler nav & links


andrewoh531:
Is this the final style for the language selector?

PR #137, opened 2018-11-12
Fix links from docs to notify.gov.au

andrewoh531:


docs/src/components/header.js
```diff
@@ -8,9 +8,7 @@ import CloseIcon from './icons/close-icon'
 import coa from '../images/coa_white.svg'
 import { Wrapper, PanelProvider } from './theme'
 import { External } from './link'
-
-const notifyBaseUrl =
-  process.env.GATSBY_NOTIFY_BASE_URL || 'https://notify.gov.au'
+import { NOTIFY_BASE_URL } from './core/env-vars'
```
jonathaningram: Just `notifyBaseURL`—don't need caps because it's just a variable.

docs/src/components/footer.js
```diff
@@ -6,6 +6,7 @@ import { height } from 'styled-system'
 import { PanelProvider, Wrapper } from './theme'
 import dta from '../images/dta-wordmark-white.svg'
 import { External } from './link'
+import { NOTIFY_BASE_URL } from './core/env-vars'
```
jonathaningram: I think I'd prefer to just use the env var directly rather than import from a global file of env vars. So just

```js
const notifyBaseURL =
  process.env. NOTIFY_BASE_URL || 'https://notify.gov.au'
```

Alternatively, what you've done reduces the chance of there being a variation in the `|| 'https://notify.gov.au'` part, so I would maybe propose this as a different way of doing this.

New file `src/url.js` (or `src/core/url.js`—whatever is best):

```js
const base = process.env. NOTIFY_BASE_URL || 'https://notify.gov.au';

export const path = (pat: string) => `${base}${pat}`; // or could be smarter to make sure you don't end up with double slashes, etc.
```

Then use it like:

```jsx
import * as url from "../wherever/url";

<Link to={url.path('/support')}>
  Support
</Link>
```

Now you encapsulate the config and logic nicely. And also super easy to test this module.

Thoughts?

docs/src/components/header.js
andrewoh531: It's being used as a constant though. I've followed this convention for constants, not sure if the team wants to adhere to it: https://www.w3schools.com/js/js_conventions.asp

docs/src/components/footer.js
jonathaningram: @andrewoh531 I'm just reading over my suggestion again. In my example I used React Router links which just need a path e.g. `/foo/bar`. In our code we have cases where we need full URLs. So how about this instead:

New file `src/url.js` (or `src/core/url.js`—whatever is best):

```js
const base = process.env. NOTIFY_BASE_URL || 'https://notify.gov.au';

const url = pat => `${base}${pat}`;

export default url;
```

Then use it like:

```jsx
import url, { path } from "../wherever/url"; // imported path just so show that it can be there if we need it too—can't recall if we need it in docs yet

<Link to={url('/support')}>
  Support
</Link>
```

docs/src/components/header.js
jonathaningram: Nope don't care about that and personally think it's often bad guidance. `notifyBaseURL` comes from an env var at the moment but it could come from runtime config. At the end of the day it's just a variable. Also by having only one use of it in url.js, it is easy to change anyway so no big deal I guess. Prefer `notifyBaseURL`. Even things like `MAX_SIZE`—just do `const maxSize = 10` and you're set.

docs/Makefile
```diff
@@ -66,4 +69,5 @@ check-vulnerabilities:
 	help clean clean-full check-vulnerabilities \
 	install install-dev \
 	fmt test run \
-	deploy deploy-dev
+	deploy deploy-dev \
+	.env.production
```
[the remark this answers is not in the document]
andrewoh531: You're right doesn't need to be phony

docs/src/components/core/notify-url.test.js
```diff
@@ -0,0 +1,33 @@
+import notifyUrl, { stripTrailingSlash } from './notify-url'
+
+test.each`
```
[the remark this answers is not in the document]
andrewoh531: If you're curious: https://jestjs.io/docs/en/api#testeachtable-name-fn-timeout

docs/src/components/core/notify-url.js
```diff
@@ -0,0 +1,18 @@
+const baseUrl = process.env.NOTIFY_BASE_URL || 'https://notify.gov.au'
+
+const stripLeadingSlash = input => {
```
jonathaningram: Can just shorten to:

```
return (input || '').replace(/^[\/]*/, '')
```

docs/src/components/core/notify-url.js
```diff
@@ -0,0 +1,18 @@
...
+  if (typeof input === 'string') return input.replace(/^[\/]*/, '')
+
+  return input
+}
+
+const stripTrailingSlash = input => {
+  if (typeof input === 'string') return input.replace(/[\/]*$/, '')
```
jonathaningram: Same, shorten here.

docs/src/components/footer.js
```diff
@@ -50,27 +51,27 @@ const UnorderedList = ({ list }) => (
 )
 
 const firstColumn = [
-  { text: 'Support', href: 'https://notify.gov.au/support' },
+  { text: 'Support', href: notifyUrl('support') },
```
jonathaningram: I think this should be `notifyUrl('/support')` with the leading slash. One so it's obvious that it's a path (readability and findability [e.g. search for `/support` URLs]) and two because it's a path which I think should start with a slash.

docs/src/components/header.js
```diff
@@ -174,7 +172,7 @@ export default () => (
     <Root>
       <Wrapper>
         <Flex justifyContent="space-between">
-          <BannerLink href={notifyBaseUrl}>
+          <BannerLink href={notifyUrl('')}>
```
jonathaningram: Especially here because `notifyUrl('')` doesn't really mean anything whereas `notifyUrl('/')` is more obvious that it links to the root.

docs/src/components/panel.js
```diff
@@ -21,7 +21,7 @@ const Summary = styled.summary`
     content: '↓ more';
   }
 
-  ${Details}[open] &:after {
+  ${Details} [open] &:after {
```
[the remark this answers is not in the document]
andrewoh531: I'll undo this

docs/Makefile
```diff
@@ -38,11 +36,16 @@ fmt:
...
 
-public/index.html: $(SRC)
+build: public/index.html
+	
+public/index.html: $(SRC) .env.production
 	$(RUNNER) build
 
+.env.production: NOTIFY_BASE_URL ?= https://notify-$(STG).apps.y.cld.gov.au
+.env.production:
+	echo NOTIFY_BASE_URL=$(NOTIFY_BASE_URL) >> .env.production
```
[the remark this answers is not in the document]
andrewoh531: I had issues using `GATSBY_` but think I might've figured it out so don't need `.env.production`. But FYI picks up environment variables from `.env.production` when you `gatsby build` as it's run in production mode and `.env.development` when you `gatsby develop` - https://www.gatsbyjs.org/docs/environment-variables/

docs/Makefile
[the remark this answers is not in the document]
andrewoh531: The target's been removed altogether

PR #138, opened 2018-11-14
fix(ui): add vertical space between home buttons

jonathaningram:


admin/app/assets/stylesheets/views/product-page.scss
```diff
@@ -24,6 +24,14 @@
 
     .button-container {
       line-height: 45px;
+
+      .button + .button {
+        margin-top: 10px;
```
andrewoh531: Should we use `rem` here?

jonathaningram: I just copied what the left/right spacing used which was pixels. Didn't want to add rems to this section because they used pixels already.

PR #139, opened 2018-11-15
Modifying open graph metadata

andrewoh531:


admin/app/templates/govau_template.html
```diff
@@ -20,7 +20,12 @@
     <meta name="theme-color" content="#0b0c0c" />
 
     <meta name="viewport" content="width=device-width, initial-scale=1">
-    <meta property="og:image" content="{{ asset_path }}images/opengraph-image.png?0.23.0">
+
+    <meta property="og:type" content="website">
+    <meta property="og:url" content="https://notify.gov.au/">
+    <meta property="og:title" content="Notify.gov.au">
+    <meta property="og:description" content="Notification platform for Australian government agencies">
```
jonathaningram: Should this description be the same as our standard meta one:

```
Notify.gov.au lets you send emails and text messages to your users. Try it now if you work in Australian local, state or federal government.
```

?

PR #144, opened 2018-11-20
Adding new-relic configuration

andrewoh531:


.circleci/config.yml
```diff
@@ -106,6 +106,7 @@ jobs:
       - *attach_workspace
       - run: make api.install
       - run: make api.build
+      - run: make -C api newrelic.ini
```
jonathaningram: Do we mean something like `make api.setup` above `api.install`? And inside that, it would write out the ini file.

api/Makefile
```diff
@@ -107,12 +107,15 @@ create-docker-test-db:
 run: version-file ## Run a development server (reloading)
 	$(RUNNER) flask run -p $(PORT)
 
+newrelic.ini:
+	$(RUNNER) newrelic-admin generate-config $(NEW_RELIC_LICENSE_KEY) newrelic.ini
```
jonathaningram: Do they happen to have a default for the ini file or do we have to specify everywhere `newrelic.ini`?

api/manifest.yml
```diff
@@ -13,6 +13,7 @@ defaults: &defaults
     - smtp
   env:
     RUNNER: "./vcapenvwrapper --root notify-shared --root notify-api --root notify-celery --root aws --root telstra --root twilio --root smtp"
+    NEW_RELIC_CMD: "NEW_RELIC_CONFIG_FILE=newrelic.ini newrelic-admin run-program"
```
jonathaningram: Do they happen to have a default for the ini file or do we have to specify everywhere `newrelic.ini`?

.circleci/config.yml
jonathaningram: Could write out the ini file in `make api.install` but I added `api.setup` as it seems more of a setup thing.

api/Makefile
andrewoh531: There doesn't seem to be a default, without the file name it just prints it out to stdout.

api/manifest.yml
andrewoh531: There doesn't seem to be a default, without the file name it just prints it out to stdout.

.circleci/config.yml
jonathaningram: > but I don't want it to generate the newrelic.ini on developer machines.

Fair.

Yeah we probably would revert it and then either add it back it for further testing, or work out how to enable it for certain envs if we want to keep it in.

PR #145, opened 2018-11-20
docs content

dagoof:


dagoof:
sorry. it's big, and not finished. want to get this merged and will continue content in subsequent PRs.

docs/src/components/table-of-contents.js
```diff
@@ -15,10 +15,18 @@ const Contents = styled.section`
 `
 
 const Ul = styled.ul`
+  /*
```
jonathaningram: Just delete this if you don't want it? OK for now though.

docs/src/pages/index.js
```diff
@@ -1,43 +1,34 @@
...
-        <li>
-          <Link
-            blank
-            href="https://github.com/alphagov/notifications-node-client"
-          >
-            Node.js
-          </Link>
-        </li>
-      </ul>
+      <div className="prevent-next-paragraphs-from-getting-hero-text" />
```
jonathaningram: Random `prevent-next-paragraphs-from-getting-hero-text`?

dagoof: we've got a pretty hectic adjacent sibling selector going on here https://github.com/govau/notify/blob/master/docs/src/components/core/heading.js#L8

this is just a hack to prevent it from selecting all these following paragraphs; which I attempted to document with this stupid classname. does not seem to have served its intended purpose.

PR #154, opened 2018-11-23
Update prettier to support mdx

dalehurwitz:


docs/src/components/panel.js
```diff
@@ -27,7 +27,7 @@ const Summary = styled.summary`
     content: '↓ show';
   }
 
-  ${Details}[open] &:after {
+  ${Details} [open] &:after {
```
jonathaningram: @dagoof probably worth fixing this sooner rather than later? @dalehurwitz this is a prettier/styled-components bug—shouldn't change here so you'll need to manually revert this one...

docs/package.json
```diff
@@ -67,5 +69,17 @@
...
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
+  "lint-staged": {
+    "*.{js,json,jsx,md,mdx,ts,tsx}": [
+      "npm run fmt",
+      "npm run fmt-check",
```
jonathaningram: Do we need this? `fmt-check` was/is just meant to be a way to check in CI if things were formatted (if we want to put this in CI). But if you're running `npm run fmt` on staged things, then `npm run fmt-check` might be redundant. Thoughts?

docs/src/components/panel.js
dagoof: yep. I'll fix now

dagoof: https://github.com/govau/notify/pull/156

docs/package.json
```diff
@@ -67,5 +69,17 @@
     "setupFiles": [
       "<rootDir>/loadershim.js"
     ]
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
```
dagoof: Can we talk about husky as a team before we merge?

jonathaningram:
Yep definitely. Let’s discuss next week before you merge.

docs/package.json
dalehurwitz: Prettier couldn't format some of the `mdx` files correctly - I resolved these issues manually. I added the `fmt-check` in case we run into this issue again. Probably redundant now though as it's unlikely we will.

docs/package.json
dalehurwitz: Happy to discuss, perhaps it's overkill.

docs/src/code-examples/sending-texts/node.mdx
```diff
@@ -62,10 +60,10 @@ substituted in the message.
 For example:
 
 ```json inline
-personalisation = {
-  first_name: "Amala",
-  reference_number: "300241"
-}
+(personalisation = {
```
jonathaningram: Something's up here. Probably should be `js` not `json` lang.

docs/src/code-examples/sending-texts/python.mdx
```diff
@@ -65,25 +65,23 @@ A unique identifier you create. This reference identifies a single unique notifi
 If successful, the shape of the response you recieve will resemble the
 following.
 
-
 ```json
 {
   "content": {
-    "body": "Hey there. Lisa\n\nPlease pay us $15.50 at your earliest convenience.\n",
```
jonathaningram: Out of interest did Prettier remove the newlines inside here or you? It's fine without them (neater), but probably not great if prettier is removing them...

dalehurwitz: I removed these

PR #158, opened 2018-11-28
Remove old docs page and redirect to new docs

dalehurwitz:


admin/app/main/views/index.py
```diff
@@ -177,4 +173,13 @@ def old_page_redirects():
         'main.old_using_notify': 'main.using_notify',
         'main.information_risk_management': 'main.security',
     }
-    return redirect(url_for(redirects[request.endpoint]), code=301)
+
+    redirects_external = {
+        'main.old_documentation': current_app.config['DOCS_BASE_URL'],
```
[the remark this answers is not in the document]
dalehurwitz: Good point, are any of the redirects necessary then?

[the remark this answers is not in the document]
dalehurwitz: @dagoof I've updated with the changes we discussed

admin/app/templates/admin_template.html
```diff
@@ -57,15 +57,15 @@
           {% endif %}
           {{ top_nav_item(url_for('main.support'), 'Support') }}
           {% if current_user.is_authenticated %}
-            {{ top_nav_item(url_for('main.documentation'), 'Documentation') }}
+            {{ top_nav_item(docs_base_url, 'Documentation') }}
```
dagoof: can these just all stay as what they were?

jonathaningram: What's your concern peter? I guess I see that it would link to the actual docs website URL rather than the internal one which always hits a redirect.

dagoof: concern was having 3 ways of doing the same thing. appreciate just having one obvious way to get there but having the direct link is fine too.

approved, go go go go

PR #166, opened 2019-01-08
Fix pyyaml vulnerability

andrewoh531:
Had to add the `pyyaml` transitive-dependency as a direct dependency as `pipenv update` did not update it.

### DIFF
--- a/api/Pipfile
+++ b/api/Pipfile
@@ -46,7 +46,7 @@ pytest-xdist = "==1.22.2"
 requests-mock = "==1.4.0"
 rfc3987 = "==1.3.7"
 strict-rfc3339 = "==0.7"
-pyyaml = ">=4.2b1"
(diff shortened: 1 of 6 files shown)


api/Pipfile
```diff
@@ -46,6 +46,10 @@ pytest-xdist = "==1.22.2"
 requests-mock = "==1.4.0"
 rfc3987 = "==1.3.7"
 strict-rfc3339 = "==0.7"
+pyyaml = ">=4.2b1"
```
jonathaningram: @andrewoh531 does it matter that the other ones are `4.2b4` but this is `4.2b1`?

andrewoh531: It resolves to `4.24b` in the `Pipfile.lock` but I'll update it to be consistent with the other `Pipfile`s

PR #178, opened 2019-04-01
Add google analytics to docs

andrewoh531:


docs/gatsby-config.js
```diff
@@ -16,6 +16,21 @@ module.exports = {
...
       },
     },
+    {
+      resolve: `gatsby-plugin-google-analytics`,
+      options: {
+        trackingId: "UA-61222473-21",
+        // Puts tracking script in the head instead of the body
+        head: false,
+        // Setting this parameter is optional
+        anonymize: true,
```
jonathaningram: @andrewoh531 don't think we need the `anonymize` param? https://www.gatsbyjs.org/packages/gatsby-plugin-google-analytics/#the-anonymize-option

I don't think we do this for notify.gov.au so if not, probably best to leave this off and if we add it on, we do it to both at the same time.

PR #196, opened 2019-04-18
Modify pricing page

andrewoh531:


admin/app/templates/views/pricing.html
```diff
@@ -24,7 +24,46 @@ <h2 class="heading-medium">Emails</h2>
...
+        {% endfor %}
+      {% endcall %}
+    </div>
+
+
+    <p>Note that these are current text message costs. Notify pricing may change as more telco providers are integrated onto Notify.</p>
+    <p>Customers will be notified when Notify pricing changes.</p>
+    <p>
+        Notify uses authorised Telco Providers from the 
+        <a target="_blank" href="https://www.dta.gov.au/help-and-advice/ict-procurement/tools-sourcing-digital-products-and-services/ict-panels-and-arrangements/buying-cloud-services">Cloud Services Panel.</a>
```
jonathaningram: @andrewoh531 I don't think we need target=_blank here, the user can just open in a new tab or press back if they like.

andrewoh531:
@jonathaningram are we ok to merge this or do we want to hold off on it still?

jonathaningram:
@andrewoh531 not yet, let's hold off.

PR #197, opened 2019-04-22
feat(utils): upgrade urllib3 to 1.24.2 or later

jonathaningram:


admin/Pipfile
```diff
@@ -23,6 +23,7 @@ notifications-python-client = "==4.8.1"
 notifications-utils = {path = "./../utils",editable = true}
 sentry-sdk = {extras = ["flask"]}
 itsdangerous = "==1.1.0"
+urllib3 = ">=1.24.2"
```
andrewoh531: Out of curiosity, did `pipenv update` not update this dependency?

jonathaningram:
I probably did it wrong. I’ll try again. My bad.

admin/Pipfile
jonathaningram: @andrewoh531 should be better now.

PR #199, opened 2019-05-01
Use SES for sending emails

jonathaningram:


admin/app/notify_client/complaint_api_client.py
```diff
@@ -0,0 +1,18 @@
+from app.notify_client import NotifyAdminAPIClient
+
+
+class ComplaintApiClient(NotifyAdminAPIClient):
+    # Fudge assert in the super __init__ so
+    # we can set those variables later.
+    def __init__(self):
+        super().__init__("a" * 73, "b")
```
dalehurwitz: Im curious, what are the mentioned variables?

api/app/notifications/notifications_ses_callback.py
```diff
@@ -1,105 +1,193 @@
...
-    client_name = 'SES'
-    try:
-        errors = validate_callback_data(data=ses_request, fields=['Message'], client_name=client_name)
-        if errors:
-            return errors
+register_errors(ses_callback_blueprint)
+
+
+@ses_callback_blueprint.route('/notifications/email/ses', methods=['POST'])
+def sns_callback_handler():
```
dalehurwitz: What's sns?

jonathaningram:
AWS Simple Notification Service.

jonathaningram:
The API key and secret passed to the client, I believe.

admin/app/templates/views/callbacks.html
```diff
@@ -53,9 +53,9 @@ <h2 class="heading-medium">Text messages you receive</h2>
         caption_visible=False
       ) %}
         {% for key, description, format in [
-          ('id', 'Notify’s id for the received message', 'UUID'),
+          ('id', 'Notify’s ID for the received message', 'UUID'),
           ('source_number', 'The phone number the message was sent from', '447700912345'),
```
andrewoh531: Change country code of example?

api/app/__init__.py
```diff
@@ -146,6 +155,8 @@ def register_blueprint(application):
     application.register_blueprint(status_blueprint)
 
     # delivery receipts
+    ses_callback_blueprint.before_request(requires_no_auth)
```
andrewoh531: Unless I missed it, we should verify messages sent to this endpoint are from AWS to protect ourselves from DOS attacks. It's unlikely to happen now when we're still early stages but when use picks up we should consider implementing it.

https://docs.aws.amazon.com/sns/latest/dg/SendMessageToHttp.verify.signature.html

.circleci/config.yml
```diff
@@ -29,6 +29,17 @@ references:
...
       at: .
 
+  aws_install: &aws_install
+    run:
+      command: |
+        curl "https://s3.amazonaws.com/aws-cli/awscli-bundle.zip" -o "awscli-bundle.zip"
+        unzip awscli-bundle.zip
+        ./awscli-bundle/install -b ~/bin/aws
+        ls -la ~/.local/lib/
+        ls -la $HOME/bin/
```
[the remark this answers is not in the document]
jonathaningram: Was removed, your comment was probably from an older version ⚡️

admin/app/templates/views/callbacks.html
jonathaningram: Done thanks.

api/app/__init__.py
jonathaningram: You're right. This link will give us some existing Python implementations as it doesn't look like there's any official SDK support: https://github.com/boto/boto3/issues/1469

PR #211, opened 2019-05-17
admin: hide service name from unauthorised user

dalehurwitz:


admin/app/templates/withoutnav_template.html
```diff
@@ -2,7 +2,7 @@
 
 {% block fullwidth_content %}
   <div id="content">
-    {% if current_service %}
+    {% if current_service and current_user.is_authenticated and current_user.belongs_to_service(current_service.id) %}
```
jonathaningram: @dalehurwitz I don't think `belongs_to_service` is defined? Searched the codebase, can't find it. Seems to be newly defined here: https://github.com/alphagov/notifications-admin/commit/558ae87baa806842c73db3e638e30c83745f4690 are you able to more or less just pull in the changes in that commit?

dalehurwitz: Apologies, I managed to revert this change somehow before I committed it. Fixed.

jonathaningram: @dalehurwitz all good!

PR #212, opened 2019-05-20
Update docs and api for smoke tests

andrewoh531:


api/app/config.py
```diff
@@ -306,9 +306,9 @@ class Config(object):
     SENDING_NOTIFICATIONS_TIMEOUT_PERIOD = 259200  # 3 days
 
     SIMULATED_EMAIL_ADDRESSES = (
-        'simulate-delivered@notifications.service.gov.uk',
-        'simulate-delivered-2@notifications.service.gov.uk',
-        'simulate-delivered-3@notifications.service.gov.uk',
+        'simulate-delivered@notify.digital.gov.au',
```
jonathaningram: @andrewoh531 sorry to not have told you in advance, but our sending will be from `@notify.gov.au` not `@notify.digital.gov.au` so can you change to that across the codebase? I'm aware that we have to wait for the SES PR to merge, so it's technically not correct but as master stands atm `@notify.digital.gov.au` does not exist either.

PR #217, opened 2019-05-27
Add get template listings documentation

andrewoh531:


docs/src/code-examples/sending-emails/go.mdx
```diff
@@ -12,7 +12,7 @@ options := []notify.SendEmailOption{
 resp, err := client.SendEmail(templateID, emailAddress, options...)
 ```
 
-<Panel label={<H3>Arguments</H3>}>
```
jonathaningram: @andrewoh531 as discussed, probably better to keep the semantic h3, but just render as plain.

docs/src/code-examples/generate-template-preview/node.mdx
```diff
@@ -0,0 +1,43 @@
+---
+lang: node
+---
+
+```javascript
+const personalisation = { 'foo': 'bar' };
```
jonathaningram: @andrewoh531 might want to check that all the JS stuff is running through prettier. I think prettier runs code fences inside MDX too... e.g. in this the brackets below are redundant and a bit noisy.

Also, I wonder if it's better to use `async/await` instead of `then/catch`?

andrewoh531: Thanks I've ran the prettier.

Regarding `async/await` I'm kinda tempted to keep it for brevity/readability:

```
client 
  .previewTemplateById(templateId, personalisation)
  .then(response => console.log(response))
  .catch(err => console.error(err))
```
vs
```
try {
  const resp = await client.previewTemplateById(templateId, personalisation)
  console.log(resp);
} catch(e) {
  console.error(err)
}
```
Not a big diff but just in terms of what's going on overall. Also I think most `js` devs would/should know of `async/await` and `promises`.

jonathaningram:
Makes sense. Good call.

docs/src/components/language-selector.js
```diff
@@ -83,7 +83,6 @@ const Languages = props => (
   <Context.Consumer>
     {({ current, changeLanguage }) => (
       <Fragment>
-        <H3>Select language</H3>
```
jonathaningram: @andrewoh531 any reason you removed this? Probably an accessible heading. Happy to discuss whether we need or if we should make it invisible, but maybe let's do in a later PR.

docs/src/code-examples/generate-template-preview/java.mdx
```diff
@@ -0,0 +1,40 @@
...
+);
+```
+
+The parameters in the personalisation argument must match the placeholder fields in the actual template.
+The Java client ignores any extra fields in the method.
+
+<Panel label={<H3 appearAs={H5}>Arguments</H3>}>
+  <Parameters>
+    <Parameter required name="templateId" kind="String">
+      Sign in to Notify.gov.au and go to the Templates page to find the template
```
jonathaningram: @andrewoh531 as long as most of the docs already say `Notify` instead of `Notify.gov.au`, then maybe prefer `Notify`?

docs/src/components/language-selector.js
andrewoh531: I've added it back in. I initially removed it as there were a lot of things going on on the page.

jonathaningram:
@andrewoh531 see 

image

The `Not yet available` is the first thing you see on this page and is not really that pretty and unaligned so for this PR could we at least change the Go ones to be markdown fences with a comment?

```
'''go
// Not yet available
'''
```

PR #221, opened 2019-06-03
Rename all "GOV.AU Notify" references

andrewoh531:


admin/app/assets/error_pages/5xx.html
```diff
@@ -46,7 +46,7 @@
         <div class="header-global">
           <div class="header-logo">
             <a href="/" title="Go to the GOV.AU homepage" id="logo" class="content">
```
jonathaningram: @andrewoh531 another ref

PR #228, opened 2019-06-14
api: add /notifications/email/ses for topic sub

jonathaningram:
Adding this piece allows for us to get Notify.gov.au to subscribe to the
SES SNS topic. This endpoint will autoconfirm the subscription when it
is created offline using the AWS CLI.

These changes were extracted out of the greater SES changes due to land
shortly.

### DIFF
--- a/api/app/notifications/notifications_ses_callback.py
+++ b/api/app/notifications/notifications_ses_callback.py
@@ -1,8 +1,13 @@
 from datetime import datetime
+import enum
+import requests
 
 from flask import (
+    Blueprint,
+    request,
     current_app,
-    json
+    json,
+    jsonify,
 )
 
 from app import statsd_client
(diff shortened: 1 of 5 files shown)


api/app/notifications/notifications_ses_callback.py
```diff
@@ -12,11 +17,78 @@
...
+
+
+def verify_message_type(message_type: str):
+    try:
+        SNSMessageType(message_type)
+    except ValueError:
+        raise InvalidMessageTypeException(f'{message_type} is not a valid message type.')
+
+
+certificate_cache = dict()
```
dagoof: we should use one of the flask cache systems:

https://werkzeug.palletsprojects.com/en/0.15.x/contrib/cache/#cache-systems

`simplecache` is not great but better than a dict; and can be migrated to a `rediscache` when we're able to support it

dagoof: http://flask.pocoo.org/docs/1.0/patterns/caching/

jonathaningram: @dagoof thanks for that—I'll pull in a simple cache for now.

PR #233, opened 2019-06-20
email: fix smart quotes formatting

jonathaningram:
Prior to this change, a smart quote inside an email address would cause
the HTML email link formatting to break.

With this change, smart quotes are now converted to plain quotes inside
an email address.

### DIFF
--- a/utils/notifications_utils/template.py
+++ b/utils/notifications_utils/template.py
@@ -34,6 +34,7 @@ from notifications_utils.formatters import (
     strip_leading_whitespace,
     add_trailing_newline,
     normalise_newlines,
+    remove_smart_quotes_from_email_addresses,
     strip_unsupported_characters,
 )
 from notifications_utils.take import Take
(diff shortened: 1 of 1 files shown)


utils/notifications_utils/template.py
```diff
@@ -605,6 +606,8 @@ def do_nice_typography(value):
         remove_whitespace_before_punctuation
     ).then(
         make_quotes_smart
+    ).then(
+        remove_smart_quotes_from_email_addresses
```
andrewoh531: Rename `remove` to `replace`? I thought it was actually just removing until I saw the logic.

jonathaningram: @andrewoh531 you're probably right but I'm hesitant to do it simply to avoid diverging where possible.

jonathaningram: @andrewoh531 at some stage we should make a call for how we want to approach upstream and diverging, etc. but not yet.

PR #251, opened 2019-07-04
api: add ability to archive a user (from GDS)

jonathaningram:
See https://github.com/alphagov/notifications-api/commit/87cb6f25974d394a8fa1b8f9cd7abe8aa28e05a7
See https://github.com/alphagov/notifications-api/commit/33166f3fb2825b360ab19569f08414dcf52a886a

### DIFF
--- a/api/app/dao/users_dao.py
+++ b/api/app/dao/users_dao.py
@@ -1,11 +1,16 @@
 from random import (SystemRandom)
 from datetime import (datetime, timedelta)
+import uuid
 
 from sqlalchemy import func
 from sqlalchemy.orm import joinedload
 
 from app import db
-from app.models import (User, VerifyCode)
+from app.dao.permissions_dao import permission_dao
+from app.dao.service_user_dao import dao_get_service_users_by_user_id
+from app.dao.dao_utils import transactional
+from app.errors import InvalidRequest
+from app.models import (EMAIL_AUTH_TYPE, User, VerifyCode)
 
 
 def _remove_values_for_keys_if_present(dict, keys):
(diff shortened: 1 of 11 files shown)


api/app/dao/users_dao.py
```diff
@@ -147,3 +152,49 @@ def get_user_and_accounts(user_id):
...
+
+@transactional
+def dao_archive_user(user):
+    if not user_can_be_archived(user):
+        msg = "User can’t be removed from a service - check all services have another team member with manage_settings"
+        raise InvalidRequest(msg, 400)
+
+    permission_dao.remove_user_service_permissions_for_all_services(user)
+
+    service_users = dao_get_service_users_by_user_id(user.id)
```
dagoof: surprised this logic isn't somewhere already. I guess this is the place it is

api/app/models.py
```diff
@@ -121,7 +121,7 @@ class User(db.Model):
     services = db.relationship(
         'Service',
         secondary='user_to_service',
-        backref='user_to_service')
+        backref='users')
```
dagoof: was this just a mistake? seems like this would require a few things to get cleaned up

api/app/user/rest.py
```diff
@@ -92,6 +93,14 @@ def update_user_attribute(user_id):
     return jsonify(data=user_to_update.serialize()), 200
 
 
+@user_blueprint.route('/<uuid:user_id>/archive', methods=['POST'])
+def archive_user(user_id):
```
dagoof: what's the permission necessary to perform this?

api/tests/app/conftest.py
```diff
@@ -764,33 +764,6 @@ def sample_invited_org_user(
     return create_invited_org_user(sample_organisation, sample_user)
 
 
-@pytest.fixture(scope='function')
```
dagoof: why gone

api/tests/app/db.py
```diff
@@ -663,3 +676,24 @@ def ses_notification_callback():
...
            'subscriptionArn=arn:aws:sns:eu-west-1:302763885840:preview-emails:d6aad3ef-83d6-4cf3-a470-54e2e75916da"\n}'
+
+
+def create_invited_user(service=None,
+                        to_email_address=None):
+
+    if service is None:
+        service = create_service()
+    if to_email_address is None:
+        to_email_address = 'invited_user@digital.gov.uk'
```
dagoof: gov.au

api/app/models.py
jonathaningram: Intentional. It comes off this commit: https://github.com/alphagov/notifications-api/commit/33166f3fb2825b360ab19569f08414dcf52a886a looks like just a different way of doing the joining table mapping.

api/app/user/rest.py
jonathaningram: `user_blueprint.before_request(requires_admin_auth)` —so yeah checks that it comes from admin.

api/tests/app/conftest.py
jonathaningram: refactored out in https://github.com/alphagov/notifications-api/commit/87cb6f25974d394a8fa1b8f9cd7abe8aa28e05a7#diff-aeffab693d3099ca885d19beb13c0ef5L773 —refactoring looks like a different way of creating their text fixtures. The old "def sample_x` funcs are on the way out.

api/tests/app/db.py
jonathaningram: might leave this as is for now as there's a few places in the codebase we can change this over to au, but also it may not matter.

PR #262, opened 2019-07-12
admin/scss: rework button styles to be consistent

dagoof:
fixes secondary and error buttons to use gov.au styles

makes better use of scss mixins to reduce duplication where possible


jonathaningram:
@dagoof active state:

image

jonathaningram:
@dagoof mobile view:

image

PR #274, opened 2019-07-22
api: add Telstra as a provider

jonathaningram:
Update Telstra provider to ensure we don't set callback URL if host is
empty.

In development, the Telstra notify URL can end up with an empty host,
so this ensures that it's not passed to Telstra of the host is empty.

Ensure that a subscription/number is always provisioned.

Use a different Telstra app for Notify system service.

TODO:

- [x] Need to recall why we are using a different API key for the Notify system service. Would have been for a reason...
- [x] Change the number of days that the subscription lives for from 30 to 365 (?) once we have a Telstra account that is not in trial mode.
- [x] Ensure that alphanumeric senders work. At last check, it should work but the feature is not available in trial mode.
- [x] Push to a new `feat-` branch because this one is old and the DB migration sequence has changed.
- [x] Ensure that the DB migration sequence is the right one at the time of merge (i.e. new ones haven't been added).


jonathaningram:
Has been superseded.

PR #320, opened 2020-02-24
celery: use SQS backend

jonathaningram:


api/Pipfile
```diff
@@ -5,7 +5,8 @@ name = "pypi"
 
 [packages]
 cffi = "==1.11.5"
-celery = "==3.1.25"
+celery = {extras = ["sqs"],version = "==4.4.0"}
+kombu = {git = "https://github.com/jonathaningram/kombu.git",ref = "sqs-no-list-queues"}
```
dagoof: bring under govau?

jonathaningram: we already had a `govau` fork and it just felt easier to do it there ^ but yeah should move it over.

api/app/queue_config.py
```diff
@@ -0,0 +1,68 @@
+import os
+import json
+
+
+def extract_predefined_queues():
+    url = os.environ.get('SQS_QUEUE_URL', '')
+    key = os.environ.get('SQS_AWS_ACCESS_KEY_ID', '')
+    secret = os.environ.get('SQS_AWS_SECRET_ACCESS_KEY', '')
+    if url and key and secret:
```
dagoof: Am I reading this right that we prefer local env variables to VCAP?

The way we do our other VCAP stuff is to export vars over the top of existing environment thereby taking precedence. This suggests the inverse is happening here

jonathaningram: You read it right. I will change it so that it applies VCAP values first and then overwrites if the single queue env config exists. While we do use these env as "local", there's nothing technically defining those as only for local—prod could just as well use those env vars if it made more sense to have a single queue in prod too.

jonathaningram: Actually let's chat about this.

api/migrations/env.py
```diff
@@ -16,7 +16,7 @@
 # from myapp import mymodel
 # target_metadata = mymodel.Base.metadata
 from flask import current_app
-config.set_main_option('sqlalchemy.url', current_app.config.get('SQLALCHEMY_DATABASE_URI'))
+config.set_main_option('sqlalchemy.url', current_app.config.get('SQLALCHEMY_DATABASE_URI').replace("%", "%%"))
```
dagoof: what's going on here?

jonathaningram: python dance because of special characters in the RDS password. `%` being a python formatter thing was causing an error (would have to go back to get specific error). but then later the `%` dance needs to be undone before configuring the migration context:

```python
url = config.get_main_option("sqlalchemy.url").replace("%%", "%")
context.configure(url=url)
```

dagoof: so sqlalchemy is doing something wrong with %? It's not something we can get around by using a different type of string formatting?

jonathaningram: i think it's expected behaviour: https://alembic.sqlalchemy.org/en/latest/api/config.html#alembic.config.Config.set_main_option

> A raw percent sign not part of an interpolation symbol must therefore be escaped, e.g. %%.

PR #330, opened 2020-03-02
api: ignore IntegrityError in celery task handler

jonathaningram:
SQS is at-least-once message delivery which means these celery
task handlers may fire more than once. If there is a duplicate
key IntegrityError it means that the task message was
delivered more than once.


jonathaningram:
I closed this because CF doesn't like long service instance names due to my long branch name. See #332

PR #341, opened 2020-03-19
admin: show monthly pricing in Usage page

acodegirl:


admin/app/templates/views/usage-with-letters.html
```diff
@@ -30,15 +30,13 @@ <h2 class='heading-small'>Text messages</h2>
             {{ big_number(sms_sent, 'sent', smaller=True) }}
             {{ big_number(sms_free_allowance, 'free allowance', smaller=True) }}
             {{ big_number(sms_allowance_remaining, 'free allowance remaining', smaller=True) }}
-            {#
               {% if sms_chargeable %}
                 {{ big_number(
                   sms_chargeable,
-                  'at {:.2f} cents per message'.format(sms_rate * 100),
+                  'at {:.3g}cents per message (excl. GST)'.format(sms_rate * 100),
```
jonathaningram: @acodegirl per https://guides.service.gov.au/content-guide/numbers-measurements/#measurement-and-weights “Abbreviate measures when using a number. Do not put a space between number and measure.” did you mean `c` instead of `cents`?

api/migrations/versions/0237_update_sms_rates.py
```diff
@@ -0,0 +1,31 @@
...
+    op.get_bind()
+    op.execute("UPDATE provider_rates SET rate = 6.0 WHERE provider_id = (SELECT id FROM provider_details WHERE identifier = 'telstra')")
+    op.execute("UPDATE provider_rates SET rate = 6.0 WHERE provider_id = (SELECT id FROM provider_details WHERE identifier = 'twilio')")
+    op.execute("UPDATE rates SET rate = 0.06 WHERE valid_from = '2017-06-30 14:00:00'")
+    op.execute("UPDATE rates SET rate = 0.06 WHERE valid_from = '2016-05-18 00:00:00'")
+
+
+def downgrade():
+    op.get_bind()
+    op.execute("UPDATE provider_rates SET rate = 1.8 WHERE provider_id = (SELECT id FROM provider_details WHERE identifier = 'telstra')")
```
jonathaningram: @acodegirl it doesn't really matter for these downgrade steps, but these 4 SQL commands should be run in reverse order from those in the upgrade steps.

acodegirl: Addressed. 
I noticed in many other scripts we are executing downgrade's SQL commands in the same order as upgrade.

admin/app/templates/views/usage-with-letters.html
acodegirl: replaced cents with c

api/migrations/versions/0237_update_sms_rates.py
jonathaningram: @acodegirl yeah it's all a bit messy really. it's probably just good practice so thought i would mention it :)

acodegirl: @jonathaningram It's a good learning for me. thanks for letting me know.  :)

jonathaningram:
Replaced by https://github.com/govau/notify/pull/344 because this branch name is too long for CF SQS service names...

PR #346, opened 2020-03-22
sap: initial commit of new provider

jonathaningram:


api/app/notifications/notifications_sms_callback.py
```diff
@@ -5,11 +5,38 @@
...
 
 sms_callback_blueprint = Blueprint("sms_callback", __name__, url_prefix="/notifications/sms")
 register_errors(sms_callback_blueprint)
 
 
+@sms_callback_blueprint.route('/sap/<notification_id>', methods=['POST'])
+@require_oauth(scope=None)
+def process_sap_response(notification_id):
+    client_name = 'SAP'
+    data = json.loads(request.data)
```
dagoof: https://flask.palletsprojects.com/en/1.1.x/api/#flask.Request.get_json

api/app/notifications/receive_notifications.py
```diff
@@ -11,132 +14,108 @@
...
+    response = MessagingResponse()
 
-    if not auth:
-        current_app.logger.warning("Inbound sms (MMG) no auth header")
-        abort(401)
-    elif auth.username not in current_app.config['MMG_INBOUND_SMS_USERNAME'] \
-            or auth.password not in current_app.config['MMG_INBOUND_SMS_AUTH']:
-        current_app.logger.warning("Inbound sms (MMG) incorrect username ({}) or password".format(auth.username))
-        abort(403)
+    data = json.loads(request.data)
```
dagoof: same. request.get_json

api/app/notifications/receive_notifications.py
```diff
@@ -11,132 +14,108 @@
...
 
-    inbound_number = strip_leading_sixty_one(post_data['Number'])
+    service = fetch_potential_service(data['originatingAddress'], 'sap')
 
-    service = fetch_potential_service(inbound_number, 'mmg')
     if not service:
-        # since this is an issue with our service <-> number mapping, or no inbound_sms service permission
-        # we should still tell MMG that we received it successfully
-        return 'RECEIVED', 200
+        # Since this is an issue with our service <-> number mapping, or no
```
dagoof: we should do something here to log or alert ourselves that something is going on..

could we manually tell sentry to report an error?

https://docs.sentry.io/error-reporting/capturing/?platform=python#capturing-messages

api/migrations/versions/0238_add_sap_sms_provider.py
```diff
@@ -0,0 +1,45 @@
...
+down_revision = '0237'
+
+
+def upgrade():
+    op.execute(
+        "INSERT INTO provider_details (id, display_name, identifier, priority, notification_type, active, supports_international, version) values ('{}', 'SAP', 'sap', 20, 'sms', true, true, 1)".format(str(uuid.uuid4()))
+    )
+
+    op.execute((
+        "INSERT INTO provider_rates (id, valid_from, rate, provider_id) VALUES ('{}', '{}', 4.0, "
```
dagoof: we want this as 6.0

jonathaningram: updated locally. won't fiddle with the branch or DB migrations for this one, so expect the price is wrong in this feature env

api/scripts/export-ngrok.py
```diff
@@ -10,7 +10,7 @@
         if tunnel['proto'] != 'https':
             continue
 
-        if tunnel['config']['addr'] != 'localhost:6011':
+        if tunnel['config']['addr'] != 'http://localhost:6011':
```
dagoof: has ngrok changed? this is hostname without schema for me.

also this will never resolve because we've skipped unless https in the check above

api/app/notifications/notifications_sms_callback.py
jonathaningram: this change would expect that SAP is sending us correct Content-Type headers, otherwise it will return None. maybe they are, do we want to check?

api/scripts/export-ngrok.py
jonathaningram: i think it changed because it didn't work for me unless i made this change.

with this change in place:

```
$ ngrok --version
ngrok version 2.3.35
$ pipenv run python scripts/export-ngrok.py
Loading .env environment variables…
https://redacted.ngrok.io
```

if i revert this change:

```
$ pipenv run python scripts/export-ngrok.py
Loading .env environment variables…
```

i.e. no output

api/app/notifications/notifications_sms_callback.py
jonathaningram: i just changed this. we can just test this in environment and see that it works. it most likely has the right header anyway. there's two other places in the codebase this is used, but i'm not changing them now (one for telstra and one for SNS)

PR #349, opened 2020-03-23
telstra: handle when number does not exist

jonathaningram:
The Telstra client send_sms call now returns the message ID and status
to upstream callers. A later commit will use these values and store them
against the notification.


api/app/clients/sms/telstra.py
```diff
@@ -51,11 +53,31 @@ def send_sms(self, to, content, reference, sender=None):
...
             resp = telstra_api.send_sms(payload)
             message = resp.messages[0]
             self.logger.info(f"Telstra send SMS request for {reference} succeeded: {message.message_id}")
 
+            return message.message_id, NOTIFICATION_SENDING
+        except telstra.rest.ApiException as e:
+            try:
+                messages = json.loads(e.body)["messages"]
+
+                if e.status == 400 and len(messages) > 0 and messages[0]["code"] == "TO-MSISDN-NOT-VALID" and messages[0]["deliveryStatus"] == "DeliveryImpossible":
```
dagoof: we'll need to cover other error types here when possible

jonathaningram: yeah this line is a bit meh, but the alternative was to prematurely pull this out into a general function with consts, etc. but seemed overkill for now

PR #368, opened 2020-04-15
docs: add status callback option

jonathaningram:


docs/src/code-examples/sending-texts/python.mdx
```diff
@@ -63,6 +63,37 @@ If you omit this argument your default SMS sender will be set for the notificati
...
+  <Parameter optional name="status_callback_url">
+
+The status callback arguments allow you to receive delivery status updates to a server of your choice. If you do not use this argument and you have set up a delivery status callback URL on your Notify service, then that will be used.
+
+This is an HTTPS URL of the server that you would like delivery status updates to be sent to.
+
+For example:
+
+```python inline nonumbers
+status_callback_url = 'https://localhost/callback'
```
dagoof, line 75: I'm not sure if localhost is the right example to be using but I don't know what is.

I suppose ideally we'd be putting something here that won't cause problems for notify if its copy-pasted. something clearly wrong that should be changed

PR #374, opened 2020-04-27
SAP: Invalidate failed auth token

acodegirl:


api/app/clients/sms/sap.py
```diff
@@ -119,6 +119,19 @@ def send_sms(self, to, content, reference, sender=None):
...
             return order.livelink_order_id[0], NOTIFICATION_SENDING
+        except ApiException as e:
+            if e.status == 401:
+                key = 'auth'
+                auth_api = AuthorizationApi(ApiClient(Configuration(
+                    username=self._client_id,
+                    password=self._client_secret,
+                )))
+
+                resp = auth_api.token_authorization_using_post1('client_credentials')
```
dagoof, line 130: we need to try to send the sms again here; not just refresh the token

acodegirl: Addressed review comment.

api/app/clients/sms/sap.py
```diff
@@ -110,15 +110,26 @@ def send_sms(self, to, content, reference, sender=None):
...
+                access_token = resp.access_token
+                auth_cache.set(key, access_token, timeout=30 * 60)
```
dagoof, line 124: we already have logic to get this token  -- we should just lean on that existing logic

also we should remove this nested exception handling logic. we can do this by clearing the cache and re-calling this method with some flag to indicate whether we are retrying or not

acodegirl: Earlier i was not sure whether to call the same method again and not to mix up with retry functionality.  Thanks for your suggestion to use a flag to differentiate the workflow.

PR #378, opened 2020-05-06
fix broken utils testcases

acodegirl:
Fixed broken unit tests in files
- test_recipient_csv
- test_recipient_validation
- test_timezone

### DIFF
--- a/utils/tests/test_recipient_csv.py
+++ b/utils/tests/test_recipient_csv.py
@@ -389,9 +389,9 @@ def test_get_recipient_respects_order(file_contents,
         (
             """
                 phone number,name
-                07800900460,test1
-                07800900460,test1
-                07800900460,test1
+                0780090046,test1
+                0780090047,test1
+                0780090048,test1
             """,
             'sms',
             ['phone number', 'name'],
@@ -443,8 +443,7 @@ def test_column_headers(file_contents, template_type, expected, expected_missing
     recipients = RecipientCSV(file_contents, template_type=template_type, placeholders=['name'])
     assert recipients.column_headers == expected
     assert recipients.missing_column_headers == expected_missing
-    if expected_missing:
-        assert recipients.has_errors is True
+    assert recipients.has_errors == bool(expected_missing)
 
 
 @pytest.mark.parametrize(
--- a/utils/tests/test_recipient_csv.py
+++ b/utils/tests/test_recipient_csv.py
@@ -168,7 +168,7 @@ def test_get_rows(file_contents, template_type, expected):
             """
                 phone number,name
                 0780090046, test1
-                0780090047, test2
+                0412345678, test2
             """,
             'sms',
             [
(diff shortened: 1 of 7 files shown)


utils/tests/test_recipient_validation.py
```diff
@@ -97,7 +127,7 @@
     lambda number: number[0] not in {
         '712345678910',   # Could be Russia
     },
-    invalid_local_phone_numbers
+    invalid_international_phone_numbers
```
dagoof, line 130: seems like this changes things? is this a big deal?

acodegirl: In few cases, we have different set of error messages for local and international phone numbers. 
And also our code is different from U.K's. So using the same set of error messages is failing. 
Hence i have to add new parameters(invalid_international_phone_numbers) for  international number validation. Please refer 'validate_phone_number_and_require_local'  & 'validate_phone_number_and_allow_international' methods. 

As per our code, when local number(prefixed with 0) exceeds 10 digits it is not a valid number. This is why i had to truncate many local phone numbers in testcase parameters.

utils/tests/test_recipient_csv.py
```diff
@@ -239,21 +238,21 @@ def test_get_rows_with_errors():
...
-    ('email', 500, "email address\n", "test@example.com\n", "test at example dot com"),
-    ('sms', 500, "phone number\n", "07900900123\n", "12345"),
+@pytest.mark.parametrize('template_type, row_count, header, filler, row_with_errors', [
+    ('email', 200, "email address\n", "testexample.com\n", 200),
+    ('sms', 500, "phone number\n", "079009001\n", 500),
 ])
-def test_big_list_validates_right_through(template_type, row_count, header, filler, row_with_error):
+def test_big_list_validates_right_through(template_type, row_count, header, filler, row_with_errors):
     big_csv = RecipientCSV(
-        header + (filler * (row_count - 1) + row_with_error),
```
dagoof, line 248: this really changes what this test is doing. I don't think this should be changed

acodegirl: My bad, I misunderstood and complicated the things.  :( 
But it's a simple fix.  :) 
Fixed it and updated.

utils/tests/test_recipient_csv.py
```diff
@@ -443,7 +443,8 @@ def test_column_headers(file_contents, template_type, expected, expected_missing
     recipients = RecipientCSV(file_contents, template_type=template_type, placeholders=['name'])
     assert recipients.column_headers == expected
     assert recipients.missing_column_headers == expected_missing
-    assert recipients.has_errors == bool(expected_missing)
+    if expected_missing:
```
dagoof, line 446: this changes the test; fix the input not the check

dagoof: use this number for second set of inputs `0412 345678`

PR #393, opened 2020-05-25
Reverify email once a month

jonalport:
**Feature flag ENV variable required**
`FEATURE_EMAIL_REVERIFICATION=on`

**Screenshots**
 - https://govausites.atlassian.net/browse/NOT-265

**Tickets**
 - https://govausites.atlassian.net/browse/NOT-290
 - https://govausites.atlassian.net/browse/NOT-291
 - https://govausites.atlassian.net/browse/NOT-333


admin/app/main/views/reverify_email.py
```diff
@@ -0,0 +1,64 @@
...
+    if not user:
+        abort(404)
+
+    session['user_details'] = {
+        'email': user.email_address,
+        'id': user.id,
+        'email_reverified': True,
+    }
+
+    user_api_client.send_verify_code(user.id, 'sms', user.mobile_number)
```
dagoof, line 36: same as comment on other ticket about authentication methods

admin/app/main/views/reverify_email.py
```diff
@@ -0,0 +1,64 @@
...
+from flask import request, render_template, session, abort, current_app, flash, redirect, url_for
+from flask_login import current_user
+from notifications_utils.url_safe_token import check_token
+from itsdangerous import SignatureExpired
+from app import user_api_client
+from app.main import main
+from app.utils import redirect_to_sign_in
+
+
+@main.route('/reverify-email-token/<token>')
```
dagoof, line 11: are we able to use the existing 'new-password' handler for this?

admin/app/main/views/reverify_email.py
jonalport: I've updated this to exclude email auth users from the email reverification check - my thinking is that each time they sign in with email 2fa they are already reverifying. Does that make sense?

I have also made sure that the email_last_verified_at field is also being updated during email 2fa.

jonalport:
**Updated flow**

image

image

image

image

PR #395, opened 2020-05-26
Rotate password once a year

jonalport:
**Feature flag ENV variable required**
`FEATURE_PASSWORD_ROTATION=on`

**Screenshots**
 - https://govausites.atlassian.net/browse/NOT-143?focusedCommentId=76283

**Tickets**
 - https://govausites.atlassian.net/browse/NOT-287
 - https://govausites.atlassian.net/browse/NOT-288
 - https://govausites.atlassian.net/browse/NOT-318
 - https://govausites.atlassian.net/browse/NOT-331


admin/app/main/views/rotate_password.py
```diff
@@ -0,0 +1,31 @@
...
+from flask import render_template, session, url_for, redirect
+from flask_login import current_user
+
+from app import user_api_client
+from app.main import main
+from app.main.forms import RotatePasswordForm
+from app.utils import redirect_to_sign_in
+
+
+@main.route('/change-password', methods=['GET', 'POST'])
```
dagoof, line 11: I'm now realising that the design doesn't call for it; but maybe we should prompt this through an email. see ticket for details so team can discuss. `NOT-143`

jonalport: Thanks @dagoof - have commented in ticket.

admin/app/main/views/rotate_password.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def _check_if_new_password_is_same_as_current_password(new_password):
+        return user_api_client.verify_password(session['user_details']['id'], new_password)
+
+    form = RotatePasswordForm(_check_if_new_password_is_same_as_current_password)
+
+    if form.validate_on_submit():
+        new_password = form.new_password.data
+        session['user_details']['password'] = new_password
+        user_api_client.send_verify_code(session['user_details']['id'], 'sms', session['user_details']['mobile_number'])
```
dagoof, line 23: this might not always be correct. we allow some users to verify by email. see line 57/59 in admin/app/main/views/sign_in.py 

we probably need to duplicate that logic or preferably use some common mechanism to keep them in sync

jonalport: Ah yes, I will take a look - thanks

jonalport: This is actually a bit of a problem. I have added in the email 2fa bit. However, if an email auth user updates their password during this process, and then subsequently opens the email auth link in a different session/browser, then the password update will not be applied.

I feel like there are probably not many email auth users, and probably not many of those will then open the link in a different session/browser, however a bug nonetheless.

Best solution I can think of is two move the 2-factor auth before the password setting.

jonalport:
**Updated flow**

image

image

image

image

jonalport:
**Email-based flow (following link in private browsing window)**

image

image

image

image

PR #398, opened 2020-05-27
Twilio decommission

acodegirl:
Deactivated Twilio from providers list.

### DIFF
--- a/api/app/dao/provider_details_dao.py
+++ b/api/app/dao/provider_details_dao.py
@@ -27,19 +27,10 @@ def get_provider_details_by_identifier(identifier):
 
 def get_alternative_sms_provider(identifier):
     alternate_provider = None
-    if identifier == 'telstra':
+    if identifier != 'sap':
         alternate_provider = 'sap'
     else:
-        if os.getenv('FEATURE_SAP_ENABLED'):
-            if identifier == 'sap':
-                alternate_provider = 'telstra'
-            # We need to have a third check here because our tests expect that
-            # Twilio is enabled.
-            elif identifier == 'twilio':
-                alternate_provider = 'sap'
-        else:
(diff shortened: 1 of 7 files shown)


api/app/dao/provider_details_dao.py
```diff
@@ -27,20 +27,19 @@ def get_provider_details_by_identifier(identifier):
...
-    if os.getenv('FEATURE_SAP_ENABLED'):
-        if identifier == 'sap':
-            alternate_provider = 'telstra'
-        elif identifier == 'telstra':
-            alternate_provider = 'sap'
-        # We need to have a third check here because our tests expect that
-        # Twilio is enabled.
-        elif identifier == 'twilio':
-            alternate_provider = 'sap'
+    if identifier == 'telstra':
```
dagoof, line 30: can we change the logic in here a bit? something like:

```
if identifier != 'sap':
  alternate_provider = 'sap'
else:
  alternate_provider = 'telstra'
```

acodegirl: I was not sure whether to remove Twilio case here. I will change the logic and update.

api/app/dao/provider_details_dao.py
```diff
@@ -27,20 +26,10 @@ def get_provider_details_by_identifier(identifier):
 
 def get_alternative_sms_provider(identifier):
     alternate_provider = None
-    if os.getenv('FEATURE_SAP_ENABLED'):
```
dagoof, line 30: one last thing -- can we remove references to `FEATURE_SAP_ENABLED` from `api/manifest.yml` and `test_provider_details_dao`

acodegirl: done

PR #402, opened 2020-06-03
Allow removal of text message senders via UI

acodegirl:


api/migrations/versions/0244_add_sms_sender_archived_state.py
```diff
@@ -0,0 +1,22 @@
+"""
+
+Revision ID: 0244
```
jonalport, line 3: My PR #393 also adds a migration 244 😛

acodegirl: @jonalport Thanks for letting me know. I will change it to 245.

jonalport:
I will pull this down and look at it in more detail.

jonalport:
https://govausites.atlassian.net/browse/NOT-14

jonalport:
Rebased onto master and now seems to be working fine.

PR #405, opened 2020-06-16
provider: allow preferred sms provider for service

dagoof:
and add secondary SAP client


jonalport:
I've tested this, seems to work great. And code LGTM! 🚀

PR #425, opened 2020-08-28
Don't create new error during log message formatting

jonalport:


jonalport:
@dagoof - this has been updated

dagoof (approved):

PR #426, opened 2020-08-31
status-checks: only check sms status in query

dagoof:
we're not able to go out and check these statuses, so don't bother
pulling them from the database yet.


jonalport (approved):

jonalport:
👏

PR #427, opened 2020-09-02
status-callbacks: pause when failing

dagoof:
throttle service callbacks if they are failing consistently until conditions improve


jonalport (approved):
🚀

dagoof:
need to add index on table here before we merge

PR #428, opened 2020-09-07
users: add csv user report for service managers

dagoof:
this largely exposes the same data as the users page but in an exportable way.

see `manage_users` for source implementation


jonalport (approved):
👏

PR #430, opened 2020-09-09
Remove status page app (#NOT-410)

jonalport:
New app can be accessed here:
https://notify-status-2.apps.b.cld.gov.au/admin

Domain needs to be switched to this app still.


dagoof (approved):
🚀

PR #436, opened 2020-09-24
status-checker: use periodic queue

dagoof:


jonalport (approved):
That's all it took? 👍

PR #442, opened 2020-11-05
Allow user to set time zone

jonalport:
image

image

image

### DIFF
--- a/admin/app/notify_client/models.py
+++ b/admin/app/notify_client/models.py
@@ -1,6 +1,6 @@
 from itertools import chain
 
-from flask import request, session
+from flask import request, session, current_app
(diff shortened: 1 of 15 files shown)


admin/app/notify_client/models.py
```diff
@@ -117,6 +118,10 @@ def is_authenticated(self):
     def permissions(self):
         return self._permissions
 
+    @property
+    def time_zone(self):
+        return self._time_zone if self._time_zone else current_app.config['DEFAULT_TIMEZONE']
```
dagoof, line 123: This bit needs to behave a bit more defensively. In our feature env my user seems to have this field returning 'false' which makes any page with a date crash

we may need to make sure that this parses as a correct time zone here and fall back to default otherwise

image

